PR #9, opened 2021-12-20
Fix benchmarking and added weights

ashi20192020:


pallets/fragments/src/weights.rs
```diff
@@ -45,86 +42,58 @@
...
-	// Storage: System ParentHash (r:1 w:0)
-	// Storage: TransactionStorage Transactions (r:1 w:0)
-	fn check_proof_max() -> Weight {
-		(182_886_000 as Weight)
-			.saturating_add(RocksDbWeight::get().reads(5 as Weight))
-			.saturating_add(RocksDbWeight::get().writes(1 as Weight))
+	// Storage: Fragments Fragments (r:1 w:1)
+	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
+	// Storage: Fragments UnverifiedFragments (r:1 w:1)
+	fn upload() -> Weight {
```
sinkingsugar, line 94: upload and also update need to be like
https://github.com/fragcolor-xyz/clamor/blob/e71d302957f3692dc38f27fe56932b35bb65760a/pallets/fragments/src/weights.rs#L64
Where you can pass the number of bytes and take that into account please.

sinkingsugar (changes requested):
Upload and update to take into account number of bytes (update has 2, mutable and immutable)

pallets/fragments/src/lib.rs
```diff
@@ -269,8 +271,7 @@ pub mod pallet {
 		}
 
 		/// Fragment upload function.
-		// TODO #1 - weight
-		#[pallet::weight(T::WeightInfo::upload())]
+		#[pallet::weight(T::WeightInfo::upload(T::MaxDataLength::get()))]
```
sinkingsugar, line 274: This should be `immutable_data.len() + mutable_data.len()`

pallets/fragments/src/lib.rs
```diff
@@ -329,9 +330,8 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Fragment upload function.
-		// TODO #1 - weight
-		#[pallet::weight(25_000)]
+		/// Fragment update function.
+		#[pallet::weight(T::WeightInfo::update())]
```
sinkingsugar, line 334: This should be `mutable_data.len()` like `upload`

pallets/fragments/src/weights.rs
```diff
@@ -60,18 +61,26 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
+		(8_000_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(1 as Weight))
 			.saturating_add(T::DbWeight::get().writes(1 as Weight))
 	}
 	// Storage: Fragments Fragments (r:1 w:1)
 	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
 	// Storage: Fragments UnverifiedFragments (r:1 w:1)
-	fn upload() -> Weight {
-		(31_000_000 as Weight)
+	fn upload(_l: u32, ) -> Weight {
```
sinkingsugar, line 71: `l` should be factored in the calculation

pallets/fragments/src/weights.rs
```diff
@@ -84,16 +93,24 @@ impl WeightInfo for () {
...
+		(8_000_000 as Weight)
 			.saturating_add(RocksDbWeight::get().reads(1 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(1 as Weight))
 	}
 	// Storage: Fragments Fragments (r:1 w:1)
 	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
 	// Storage: Fragments UnverifiedFragments (r:1 w:1)
-	fn upload() -> Weight {
-		(31_000_000 as Weight)
+	fn upload(_l: u32, ) -> Weight {
```
sinkingsugar, line 103: `l` should be factored in the calculation

pallets/fragments/src/weights.rs
```diff
@@ -84,16 +93,24 @@ impl WeightInfo for () {
...
+		(31_749_000 as Weight)
 			.saturating_add(RocksDbWeight::get().reads(3 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(2 as Weight))
 	}
+	// Storage: Fragments Fragments (r:1 w:1)
+	// Storage: Fragments DetachedFragments (r:1 w:0)
+	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
+	fn update() -> Weight {
```
sinkingsugar, line 111: this needs `l` as well like `upload`

runtime/src/lib.rs
```diff
@@ -299,11 +299,15 @@ impl pallet_sudo::Config for Runtime {
 	type Call = Call;
 }
 
+parameter_types! {
+	pub const MaxDataLength: u32 = 100;
```
sinkingsugar, line 303: I'm actually not so sure we need this because there is already another length parameter which is max extrinsic size (or max transaction size) already in the runtime core iirc. Substrate transaction storage I think mentions about it in their test/examples.

.maintain/frame-weight-template.hbs
```diff
@@ -0,0 +1,114 @@
+// This file is part of Substrate.
```
sinkingsugar, line 1: how is this used?

Cargo.lock
```diff
@@ -0,0 +1,8374 @@
+# This file is automatically @generated by Cargo.
```
sinkingsugar: I noticed you removed this file before, but now added back in.
Is that intentional?

sinkingsugar (changes requested):

pallets/fragments/src/weights.rs
sinkingsugar: This is still not factored

.maintain/frame-weight-template.hbs
ashi20192020ayush: @sinkingsugar This is weight template file and used in below command to generate weights. 

`./target/release/clamor benchmark --chain=dev --execution=wasm --wasm-execution=compiled --pallet=fragments-pallet --extrinsic='*' --steps=50 --repeat=20 --heap-pages=4096 --template=./.maintain/frame-weight-template.hbs --output=./pallets/fragments/src/weights.rs`

Cargo.lock
ashi20192020ayush: removed

pallets/fragments/src/lib.rs
sinkingsugar: Wrong, as I asked many times we need to take into account the actual data length!

pallets/fragments/src/lib.rs
sinkingsugar: we need to take into account the actual data length!

sinkingsugar (changes requested):
Need to properly take into account data len of transaction before reviewing.
For benchmarks ONLY you can use an hardcoded value of 1 megabyte

pallets/fragments/src/weights.rs
```diff
@@ -86,8 +86,10 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
 	// Storage: Fragments FragmentsListSize (r:1 w:1)
 	// Storage: Fragments FragmentsList (r:0 w:1)
-	fn upload() -> Weight {
-		(160_000_000 as Weight)
+	fn upload(l: u32, ) -> Weight {
+		(157_800_000 as Weight)
+			// Standard Error: 163_000
+			.saturating_add((1_600_000 as Weight).saturating_mul(l as Weight))
```
sinkingsugar, line 92: we are making progress but those numbers are HUGE, how can we tweak those?

ashi20192020ayush: We can do this as a part of separate issue

sinkingsugar (approved):

PR #12, opened 2022-01-02
refactor CI

sinkingsugar:


Kryptos-FR:
`.lsp` and `.clj-kondo` temp files shouldn't be included.

PR #13, opened 2022-01-02
Fixed test cases

ashi20192020ayush:


pallets/fragments/src/tests.rs
```diff
@@ -202,17 +202,29 @@ fn update_should_works() {
...
+		let signature_hash = blake2_256(&[&fragment_hash[..], &data_hash[..]].concat());
+
+		let signature: sp_core::ecdsa::Signature = pair.sign(&signature_hash);
+
+		let recover = Crypto::secp256k1_ecdsa_recover_compressed(&signature.0, &signature_hash)
+			.ok()
+			.unwrap();
+		let recover = sp_core::ecdsa::Public(recover);
+
+		assert_ok!(FragmentsPallet::add_upload_auth(Origin::root(), recover, who));
```
sinkingsugar, line 220: This doesn't test anything and it's wrong, just makes test pass, we are adding the recovered key into auth, it's wrong, we should just add `//Charlie` to auth. Nothing else to test proper signature workflow.

sinkingsugar: Please fix in the common code we talked about. 🙏

sinkingsugar (changes requested):

pallets/fragments/src/tests.rs
ashi20192020ayush: @sinkingsugar We need this to satisfy this condition https://github.com/fragcolor-xyz/clamor/blob/master/pallets/fragments/src/lib.rs#L301

sinkingsugar: @ashi20192020ayush correct but the signer is an external service off chain not the users.
So it needs to be just 1 key e.g. `//Charlie`

sinkingsugar: Users will ask to that service to verify the payload they upload is valid and receive a signature if valid, to allow them to transact basically.
Sorry this wasn't too clear.

pallets/fragments/src/tests.rs
```diff
@@ -10,6 +10,51 @@ use sp_io::{crypto as Crypto, hashing::blake2_256};
...
+	let include_info =
+		IncludeInfo { fragment_hash: FRAGMENT_HASH, mutable_index: Some(Compact(1)), staked_amount: Compact(1) };
+	let references = vec![include_info];
+
+	let pair = sp_core::ecdsa::Pair::from_string("//Alice", None).unwrap();
+
+	let fragment_hash = blake2_256(&data);
+	let signature_hash = blake2_256(&[&fragment_hash[..], &references.encode()].concat());
+
+	let who: sp_core::sr25519::Public = Default::default();
```
sinkingsugar, line 31: Make this `//Charlie`

pallets/fragments/src/tests.rs
```diff
@@ -10,6 +10,51 @@ use sp_io::{crypto as Crypto, hashing::blake2_256};
...
+	let signature: sp_core::ecdsa::Signature = pair.sign(&signature_hash);
+
+	let recover = Crypto::secp256k1_ecdsa_recover_compressed(&signature.0, &signature_hash)
```
sinkingsugar, line 34: Don't need this, recover is done inside upload, that's all we care

pallets/fragments/src/tests.rs
```diff
@@ -10,6 +10,51 @@ use sp_io::{crypto as Crypto, hashing::blake2_256};
...
+		.ok()
+		.unwrap();
+	let recover = sp_core::ecdsa::Public(recover);
+
+	assert_ok!(FragmentsPallet::add_upload_auth(Origin::root(), recover, who));
```
sinkingsugar, line 39: Just add `//Charlie` once

sinkingsugar (changes requested):

sinkingsugar (approved):

PR #24, opened 2022-01-20
Entity Pallet Implementation

ashi20192020:


pallets/entities/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+
+			ensure!(
+				!<DetachedFragments<T>>::contains_key(&fragment_hash),
+				Error::<T>::FragmentDetached
+			);
+
+			let hash = blake2_256(
+				&[
+					&fragment_hash[..],
+					&metadata.name.encode(),
```
sinkingsugar, line 133: We need to expand this a bit but for now it's good, maybe add a `//TODO` warning though

pallets/entities/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+			let entity_data = EntityData{
+				fragment_hash,
+				metadata,
+				unique,
+				mutable,
+				max_supply
+			};
+			<Entities<T>>::insert(&hash, entity_data);
+
+			Fragment2Entities::<T>::try_mutate(&fragment_hash, |entity_hash| -> DispatchResult {
```
sinkingsugar, line 152: How does this work if it was the first entity?
Shouldn't this be `mutate` and create the collection is missing? (need to double check fragments pallet too about this)

pallets/entities/src/tests.rs
```diff
@@ -0,0 +1,170 @@
...
+
+		assert_ok!(EntitiesPallet::create(
+			Origin::signed(sp_core::sr25519::Public::from_raw(PUBLIC)),
+			FRAGMENT_HASH,
+			entity_data,
+			true,
+			true,
+			None
+		));
+		assert!(Entities::<Test>::contains_key(&hash));
```
sinkingsugar, line 80: Test all the other storages being populated too

sinkingsugar (changes requested):

pallets/entities/src/lib.rs
ashi20192020ayush: I think you are talking about `mutate_exists`

You can find more details here https://github.com/paritytech/substrate/blob/monthly-2022-01/frame/support/src/storage/types/map.rs#L152

Its working fine because I am verifying through test-cases

ashi20192020ayush: I have corrected it. you can find more details here https://github.com/paritytech/substrate/blob/monthly-2022-01/frame/support/src/storage/types/map.rs#L152

pallets/entities/src/tests.rs
ashi20192020ayush: done

pallets/entities/src/lib.rs
sinkingsugar: `mutate` is enough, try is pointless here you always return `Ok(())` there is no failure path...

sinkingsugar (changes requested):

pallets/entities/src/lib.rs
sinkingsugar: Can you anyway replace this with a better structure that uses `append`? Vec is not efficient.

sinkingsugar: actually vec can stay maybe, append seems to be the same thing of this mutate.. (not 100% sure)

sinkingsugar: this is still the case though:
mutate is enough, try is pointless here you always return Ok(()) there is no failure path...

sinkingsugar (approved):

PR #39, opened 2022-03-13
Add native tokens

sinkingsugar:


pallets/protos/src/lib.rs
```diff
@@ -143,6 +156,18 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type UploadAuthorities<T: Config> = StorageValue<_, BTreeSet<ecdsa::Public>, ValueQuery>;
 
+	// Staking management
+	// (Amount staked, Last stake time)
+	#[pallet::storage]
+	pub type ProtoStakes<T: Config> = StorageDoubleMap<
```
aang114, line 162: Would it be better to use a StorageNMap instead so that we can store the BlockNumber as another key?

aang114 (approved):
The added currency (FRAG) and the code added to the pallet `protos` seems correct

**(I didn't look through the mock.rs and tests.rs files)**

pallets/protos/src/lib.rs
sinkingsugar: We don't really care though in this case, but for the Protos/RPCs maybe

PR #40, opened 2022-03-22
Test cases staking

aang114:
Added 3 new test cases for:

1. `stake_should_work()`
2. `stake_should_not_work_if_proto_not_found()`
3. `stake_should_not_work_if_amount_greater_than_balance()`

### DIFF
--- a/pallets/protos/src/tests.rs
+++ b/pallets/protos/src/tests.rs
@@ -42,6 +42,8 @@ fn initial_set_up_and_get_signature(
 	signature
 }
 
+/// Uploads Proto with data `pallet_protos::mock::DATA` (uploaded by `pallet_protos::mock::PUBLIC1`)
+/// @dev DO NOT CHANGE THIS FUNCTION
 fn initial_upload_and_get_signature() -> AuthData {
 	let data = DATA.as_bytes().to_vec();
 	let references = vec![];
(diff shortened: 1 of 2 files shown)


pallets/protos/src/tests.rs
```diff
@@ -484,3 +505,74 @@ fn internal_finalize_detach_should_works() {
 		assert!(<DetachedHashes<Test>>::contains_key(PROTO_HASH));
 	});
 }
+
+
+#[test]
+fn stake_should_work() {
```
aang114, line 511: Giovanni (@sinkingsugar), staking failed here even though the Amount Staked by Account `PUBLIC1` is equal to the Balance of Account `PUBLIC1`. 

It only worked when I made it greater (i.e when the balance / minted amount is greater than the staked amount). 

But from the pallet's code, it should work when it's **greater than or equal to.**

So there may be a bug in the pallet's code?

sinkingsugar: No bug, It's the `transfer`'s `keep_alive` parameter.
If you transfer ALL it deletes the account basically.
(I'm not 10000% sure about the utility of keeping the "account" to be honest)

sinkingsugar: Could you dig it a bit please?

sinkingsugar: updates?

sinkingsugar:
What is left to do here?

sinkingsugar:
oops CI to fix

pallets/protos/src/tests.rs
aang114: If the account is kept alive, tokens can still be sent to its address.

Polkadot gives the user the option keep their account alive when they transfer: (https://wiki.polkadot.network/docs/learn-balance-transfers#keep-alive-checks)

> By default, Polkadot-JS Apps will use transfer_keep_alive, ensuring that the account you send from cannot drop below the existential deposit of 1 DOT.
> 
> However, it may be that you do not want to keep this account alive (for example, because you are moving all of your funds to a different address). In this case, click on the "keep-alive" toggle at the bottom of the modal window. The label should switch from "Transfer with account keep-alive checks" - transfer_keep_alive will be used, to "Normal transfer without keep-alive checks" - transfer extrinsic will be used. As a common use case for using normal transfers is to entirely clear out the account, a second toggle will appear if you have the keep-alive check turned off that will send all the tokens in the account, minus a transaction fee, to the destination address.

sinkingsugar: Nice find thanks

sinkingsugar:
@aang114 I completely reworked staking in my current PR so this needs some fixing after that is merged.

PR #42, opened 2022-03-27
eth sync preps

sinkingsugar:


pallets/protos/Cargo.toml
```diff
@@ -12,32 +12,32 @@ targets = ['x86_64-unknown-linux-gnu']
 
 [dependencies.sp-core]
 default-features = false
-git = 'https://github.com/paritytech/substrate.git'
-tag = 'monthly-2022-03'
+git = 'https://github.com/clamor-io/substrate.git'
```
aang114, line 15: https://github.com/clamor-io/substrate.git is a fork of Substrate that is obviously owned by us, yes?

pallets/protos/src/lib.rs
```diff
@@ -102,6 +120,7 @@ pub mod pallet {
 	#[pallet::config]
 	pub trait Config:
 		frame_system::Config
+		+ CreateSignedTransaction<Call<Self>>
```
aang114: Note to self: Investigate what this does

pallets/protos/src/lib.rs
```diff
@@ -90,6 +95,19 @@ pub struct Proto<TAccountId, TBlockNumber> {
 	pub metadata: BTreeMap<Vec<u8>, Hash256>,
 }
 
+#[derive(Encode, Decode, Clone, scale_info::TypeInfo, Debug, PartialEq, Eq)]
+pub struct EthStakeUpdate<TPublic, TBalance> {
```
aang114: Why is `EthStakeUpdate` defined in pallet_protos and pallet_frag?

pallets/protos/src/lib.rs
```diff
@@ -722,13 +741,6 @@ pub mod pallet {
 		}
 	}
 
-	#[pallet::hooks]
-	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_finalize(_n: T::BlockNumber) {}
```
aang114, line 727: Note to self: Find out why this was removed

primitives/chainblocks/src/lib.rs
```diff
@@ -100,3 +102,48 @@ pub fn init() {
...
+
+	offchain::http_request_add_header(request, "Content-Type", "application/json")
+		.map_err(|_| "Failed to add header")?;
+
+	offchain::http_request_write_body(request, body, None).map_err(|_| "Failed to write body")?;
+
+	// send off the request
+	offchain::http_request_write_body(request, &[], None).unwrap();
+
+	let results = offchain::http_response_wait(&[request], None);
```
aang114: Giovanni, since "there are uncertainty in terms of how long the request will take, and the result may not be the same all the time. This causes problem for the network reaching consensus." (https://substrate.recipes/off-chain-workers/http-json.html)

Would it be wise if the deadline parameter was not `None` ?

aang114: Note to Self: Find out how HTTP requests can cause problem for the network reaching consensus

pallets/frag/src/weights.rs
```diff
@@ -0,0 +1,160 @@
+// This file is part of Substrate.
```
aang114, line 1: How did you create the `weights.rs` file if the `benchmark.rs` file hasn't even been created?

pallets/protos/Cargo.toml
sinkingsugar: Sort of, we are outsourcing that part but yeah we control.

pallets/protos/src/lib.rs
sinkingsugar: If you want to send transactions from a pallet you need such trait (tl;dr)

pallets/protos/src/lib.rs
sinkingsugar: Mistake, good catch.

pallets/protos/src/lib.rs
sinkingsugar: Not used I think?

primitives/chainblocks/src/lib.rs
sinkingsugar: They run in different threads/contexts

pallets/frag/src/weights.rs
sinkingsugar: This is autogenerated

sinkingsugar: I will ask you do actually work on that, so be good if you can get up to speed to the cause ;)

pallets/detach/src/lib.rs
```diff
@@ -371,7 +365,7 @@ pub mod pallet {
 				);
 				for authority in authorities {
 					<EthereumAuthorities<T>>::mutate(|authorities| {
-						authorities.insert(authority.clone());
+						authorities.insert(*authority);
```
aang114, line 368: Explanation to Self: `ecds::Public` implements the `Copy` trait, that's why this is possible

chains/common.edn
```diff
@@ -154,7 +154,6 @@
 
   "0xC0FFEE" (HexToBytes) = .patch-data-bytes
   [.patch-data-bytes] (Substrate.Encode [nil]) >= .patch-data
-  "0x01" (HexToBytes) (PrependTo .patch-data)
```
aang114, line 157: Why was this even prepended before btw?

chains/fragments-tests.edn
```diff
@@ -10,6 +10,16 @@
...
+  (ForEach
+   (->
+    (Log "connecting")
+    = .p2p-address
+    "swarm" >> .ipfs-args
+    "peering" >> .ipfs-args
+    "add" >> .ipfs-args
+    .p2p-address >> .ipfs-args
+    (Maybe (-> "" (Process.Run "ipfs" .ipfs-args)))))
+  ; also call connect
```
aang114, line 22: From my experience with using the Terminal

Once you call `ipfs swam peering add <peer-address>`, `ipfs cat <cid>` and `ipfs block get <cid>` should already work.

So I don't understand why you are calling `ipfs swarm connect <peer-address>` below (see line 34)

sinkingsugar: That was the only way to fix this situation, so there must be a reason :)

pallets/frag/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+		type WeightInfo: WeightInfo;
+
+		#[pallet::constant]
+		type FragToken: Get<<Self as pallet_assets::Config>::AssetId>;
+
+		#[pallet::constant]
+		type EthChainId: Get<u64>;
+
+		/// The identifier type for an offchain worker.
+		type AuthorityId: AppCrypto<Self::Public, Self::Signature>;
```
aang114, line 135: Explanation to Self: This tell the runtime that this pallet can create signed transactions. 
(https://substrate.recipes/off-chain-workers/transactions.html)

pallets/frag/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+			origin: OriginFor<T>,
+			signature: ecdsa::Signature,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let message = b"TODO";
+			let signature_hash = blake2_256(message);
+
+			// ! Notice we do this all in WASM... SLOWNESS here can be a issue
+			let sig = libsecp256k1::Signature::parse_overflowing_slice(&signature.0[..64])
```
aang114, line 197: What exactly are you doing here? Thanks

aang114: What exactly are you doing here? Thanks

PR #46, opened 2022-04-05
Replace issue templates

Kryptos-FR:
Replaced old templates with ones we use in our public repositories.
Removed "question" template.


Kryptos-FR:
@sinkingsugar in this repo, the old templates inherited from the fork are replaced with the one we have in other repo. Is that ok?

sinkingsugar:
I think we should replace yeah!

PR #58, opened 2022-05-09
refactor rpcs and add tags + cats

sinkingsugar:


sinkingsugar:
cc @paulinechi , @aang114 
I just need to figure how to fix the JS code now...

PR #69, opened 2022-06-04
getProtos work

sinkingsugar:
Return inclusion info in getProtos, remove a lot of complexity and O(n) from getProtos

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -932,10 +932,9 @@ pub mod pallet {
 								Value::Null
 							};
 
-							(*map_proto).insert(
-								String::from_utf8(metadata_key.clone()).unwrap(),
-								metadata_value,
-							);
+							if let Ok(key) = String::from_utf8(metadata_key.clone()) {
+								(*map_proto).insert(key, metadata_value);
+							}
 						}
 					}
 				}
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -28,7 +28,10 @@ pub use weights::WeightInfo;
 
 use sp_clamor::{get_locked_frag_account, Hash256};
 
-use scale_info::prelude::string::{String, ToString};
+use scale_info::prelude::{
+	format,
+	string::{String, ToString},
+};
 use serde_json::{json, Map, Value};
 
 use base58::ToBase58;
@@ -395,7 +398,7 @@ pub mod pallet {
 
 			<Protos<T>>::mutate(&proto_hash, |proto| {
 				let proto = proto.as_mut().unwrap();
-				if data.len() > 0 {
+				if !data.is_empty() {
 					// No failures from here on out
 					proto.patches.push(ProtoPatch {
 						block: current_block_number,
@@ -555,7 +558,7 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Detached a proto from this chain by emitting an event that includes a signature.
+		/// Detach a proto from this chain by emitting an event that includes a signature.
 		/// The remote target chain can attach this proto by using this signature.
 		///
 		///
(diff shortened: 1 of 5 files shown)


sinkingsugar:
@aang114 @paulinechi can we merge this?

pallets/protos/src/lib.rs
```diff
@@ -848,55 +851,91 @@ pub mod pallet {
...
-						ProtoOwner::ExternalAsset(_linked_asset) => String::from("ExternalAsset"),
+					let array_proto_id: Hash256 = if let Ok(array_proto_id) = hex::decode(proto_id)
+					{
+						if let Ok(array_proto_id) = array_proto_id.try_into() {
+							array_proto_id
+						} else {
+							continue;
+						}
+					} else {
+						continue;
```
aang114: Same as above

pallets/protos/src/lib.rs
```diff
@@ -848,55 +851,91 @@ pub mod pallet {
...
-
-					let string_owner = match owner {
-						ProtoOwner::User(account_id) => hex::encode(account_id.as_ref()), //format!("{:?}", account_id),
-						ProtoOwner::ExternalAsset(_linked_asset) => String::from("ExternalAsset"),
+					let array_proto_id: Hash256 = if let Ok(array_proto_id) = hex::decode(proto_id)
+					{
+						if let Ok(array_proto_id) = array_proto_id.try_into() {
+							array_proto_id
+						} else {
+							continue;
```
aang114: I think it makes more sense to return an error, rather than ignoring it with `continute`. Cos if this line ever gets executed, it's means our code in `get_protos` screwed up somewhere

pallets/protos/src/lib.rs
```diff
@@ -848,55 +851,91 @@ pub mod pallet {
...
-					match map_proto {
-						Value::Object(map_proto) => {
-							(*map_proto).insert(String::from("owner"), Value::String(string_owner))
-						},
-						_ => None,
+					let (owner, map_metadata, include_cost) =
+						if let Some(proto) = <Protos<T>>::get(array_proto_id) {
+							(proto.owner, proto.metadata, proto.include_cost)
+						} else {
+							continue;
```
aang114: same as above

pallets/protos/src/lib.rs
```diff
@@ -848,55 +851,91 @@ pub mod pallet {
...
+						};
+
+					let map_proto = match map_proto {
+						Value::Object(map_proto) => map_proto,
+						_ => continue,
```
aang114: same as above

aang114 (changes requested):
Everything looks good. I'm only concerned if we should use `continue` in a for loop or we should throw an error @sinkingsugar?

I personally think throwing an explicit error makes more sense.

pallets/protos/src/lib.rs
sinkingsugar: we continue because rest can be still valid data

pallets/protos/src/lib.rs
sinkingsugar: we continue because rest can be still valid data, it should really not happen but we might wanna log it at least

pallets/protos/src/lib.rs
sinkingsugar: we continue because rest can be still valid data, it should really not happen but we might wanna log it at least

pallets/protos/src/lib.rs
sinkingsugar: `proto_id` comes from user rpc request? hmm might wanna throw Error if so

pallets/protos/src/lib.rs
sinkingsugar: This is unlikely so should not really happen.. or actually could be .. let me review it

PR #82, opened 2022-06-14
fix #66

### LINKED ISSUE #66: Add the option to pass in parameter available in getProtos method
Add the option to pass in parameter **available** (true / false) in getProtos method

sinkingsugar:


sinkingsugar:
Mostly just need the ok from @paulinechi 
But triggered a few for review :)

aang114 (approved):
Looks good to me, but we also need to add an additional test in the JavaScript file. I can do that later

paulinechi:
Tested and works as expected. Added the missing parameter in GetProtosParams.

PR #93, opened 2022-07-05
Fix "Publish gh pages" workflow

aang114:
It wasn't working because in the shell script `scripts/update-gh-pages.sh`:

- There was a typo when performing `rm`
- I was using `mv` instead of `cp`


sinkingsugar (approved):
Good, make sure tests pass before merge though.

Issue #85, opened 2022-06-24
Refactor all the events


sinkingsugar:
Currently we use anonymous tuples, move to something clearer like:
```rust
MetadataSet {
	asset_id: T::AssetId,
	name: Vec<u8>,
	symbol: Vec<u8>,
	decimals: u8,
	is_frozen: bool,
	can_transfer: bool,
},
```

sinkingsugar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/clamor/85?planning-poker) with ZenHub @alebaffa @aang114

alebaffa:
@sinkingsugar Do you mean to refactor events like this

`Self::deposit_event(Event::InventoryAdded(to, class, (edition, copy)));`

to something like
`Self::deposit_event(Event::InventoryAdded(MetadataSet));`
?

sinkingsugar:
No, I don't want new types. Just named anonymous type within the Enum itself.

sinkingsugar:
Example @alebaffa https://github.com/fragcolor-xyz/protos/blob/2ae8ba0a96c6248f8f92514e43ca24eb0d6f38c1/src/traits.rs#L17-L23

alebaffa:
Fixed in PR #100 .

PR #100, opened 2022-07-12
Refactored all events (Issue #85)

alebaffa:
**Description**
Events in Clamor use anonymous tuples, for example:
```rust
Self::deposit_event(Event::InventoryAdded(to.clone(), *fragment_hash, (id, 1)));
```

This PR introduced named tuples to add more clarity, for example the above becomes:

```rust
Self::deposit_event(Event::InventoryAdded {
	account_id: to.clone(),
	fragment_hash: *fragment_hash,
	fragment_id: (id, 1),
});
```


sinkingsugar (approved):
LGTM

PR #112, opened 2022-07-20
Fixed bug in `unlink_account()` of Pallet Accounts

aang114:
Very small PR

### DIFF
--- a/pallets/accounts/src/lib.rs
+++ b/pallets/accounts/src/lib.rs
@@ -297,6 +297,8 @@ pub mod pallet {
 		AccountAlreadyLinked,
 		/// Account not linked
 		AccountNotLinked,
+		// Account linked to different account
+		DifferentAccountLinked,
 		/// Account already exists
 		AccountAlreadyExists,
 		/// Too many proxies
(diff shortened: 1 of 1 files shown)


pallets/accounts/src/lib.rs
```diff
@@ -868,8 +868,12 @@ pub mod pallet {
 		/// Unlink the **Clamor public account address `sender`** from **its linked EVM public
 		/// account address `account`**
 		fn unlink_account(sender: T::AccountId, account: H160) -> DispatchResult {
-			ensure!(<EVMLinks<T>>::contains_key(sender.clone()), Error::<T>::AccountNotLinked);
-			ensure!(<EVMLinksReverse<T>>::contains_key(account), Error::<T>::AccountNotLinked);
+			if <EVMLinks<T>>::get(sender.clone()).ok_or(Error::<T>::AccountNotLinked)? != account {
+				return Err(Error::<T>::AccountNotLinked.into());
```
sinkingsugar: Maybe we need a new kind of error?

sinkingsugar (changes requested):

pallets/accounts/src/lib.rs
alebaffa: Agree.

alebaffa (changes requested):

pallets/accounts/src/lib.rs
aang114: i've added a new kind of error. please let me know if it's okay. thanks

alebaffa (approved):
LGTM

pallets/accounts/src/lib.rs
```diff
@@ -869,10 +871,10 @@ pub mod pallet {
 		/// account address `account`**
 		fn unlink_account(sender: T::AccountId, account: H160) -> DispatchResult {
 			if <EVMLinks<T>>::get(sender.clone()).ok_or(Error::<T>::AccountNotLinked)? != account {
-				return Err(Error::<T>::AccountNotLinked.into());
+				return Err(Error::<T>::DifferentAccountLinked.into());
```
sinkingsugar, line 874: How about `AccountAlreadyLinked` 🤔

aang114: isn't `DifferentAccountLinked` more accurate?

sinkingsugar: I think it's equal , just for consistency with other errors, but doesn't matter really

sinkingsugar (approved):

pallets/accounts/src/lib.rs
aang114: okay then i'm just merging it 😅

PR #121, opened 2022-07-21
Preventing `link()` from linking to a random ethereum account

aang114:
Very Short PR

@sinkingsugar isn't this better / more foolproof?


sinkingsugar:
> Very Short PR
> 
> @sinkingsugar isn't this better / more foolproof?

Not really, key can be derived from signature we don't need to waste more block space with something we are deriving.

sinkingsugar (commented):
no point at all I'd say, an attacker can derive themself and pass right key....

PR #123, opened 2022-07-25
Added Test Cases for Pallet Fragments

aang114:


pallets/fragments/src/mock.rs
```diff
@@ -35,21 +32,31 @@ frame_support::construct_runtime!(
 		ProtosPallet: pallet_protos::{Pallet, Call, Storage, Event<T>},
 		FragmentsPallet: pallet_fragments::{Pallet, Call, Storage, Event<T>},
 		DetachPallet: pallet_detach::{Pallet, Call, Storage, Event<T>},
-		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
-		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>},
-		Accounts: pallet_accounts::{Pallet, Call, Storage, Event<T>},
+		BalancesPallet: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
```
sinkingsugar: Remove Pallet from all, didn't we decide that?

pallets/fragments/src/mock.rs
```diff
@@ -109,32 +116,40 @@ where
...
-	type ExistentialDeposit = ConstU64<1>;
+	type ExistentialDeposit = ConstU128<1>;
 	type AccountStore = System;
 	type WeightInfo = ();
 	type MaxLocks = ();
 	type MaxReserves = ();
 	type ReserveIdentifier = [u8; 8];
 }
 
+parameter_types! {
```
sinkingsugar, line 130: wow lot of changes... is that cos of the min balance testing?

pallets/fragments/src/mock.rs
aang114: yes 😅

pallets/fragments/src/mock.rs
aang114: yup. but since we reverted back to `deposit_creating` - it makes no difference 😅.

but yup, as you see in mock.rs - previously we set the associated type for `Balance` as `()` (in the implementation for `pallet_assets::Config` and `pallet_proxy::Config`) 🤦‍♀️. That's why we couldn't catch the error you had when running it as a local node

pallets/fragments/src/mock.rs
aang114: I renamed it for everything except `FragmentsPallet` - because there also exists a storage item called `Fragments` (which is imported in *tests.rs*). 

So if both Fragments (the pallet) and Fragments (the storage item) are imported in tests.rs - it would cause compiler errors

aang114:
The test cases that are failing (which you can see below and also in the CI) fall into one of 3 categories:

- related to proto detachment (all failing test cases except 3 fall into this category). I purposely added a `todo()!` in these test cases because anyways proto-fragment detachment is gonna be deprecated
  - @sinkingsugar @alebaffa  should I just put the `ignore!` macro on them then ?
  - or should we just remove the error that is thrown when a proto is detached! (i.e from `pallets/fragments/src/lib.rs` itself)?
- when buying a fragment instance using a non-native currency
  - @sinkingsugar could you advise how I can test this because right now I'm confused about how to create a non-native currency
using pallet_assets 😅 (you can read the test case to see what I'm currently doing)  
- related to `create_account()` - this is because `create_account()` isn't done yet 
  - @sinkingsugar @alebaffa  should we put the `ignore!` macro on them for now?


Thanks :)

```
failures:
    tests::buy_tests::buy_should_not_work_if_proto_is_detached
    tests::buy_tests::buy_should_not_work_if_user_has_insufficient_balance_in_pallet_assets
    tests::buy_tests::buy_should_work_if_user_has_sufficient_balance_in_pallet_assets
    tests::create_account_tests::create_account_should_work
    tests::create_tests::create_should_not_work_if_proto_is_detached
    tests::mint_tests::mint_should_not_work_if_proto_is_detached
    tests::publish_tests::publish_should_not_work_if_proto_is_detached
    tests::unpublish_tests::unpublish_should_not_work_if_proto_is_detached
```

alebaffa:
Yes, I would use the ignore macro on all of them.

sinkingsugar:
`create_account` should work 
TODOs are for the future

sinkingsugar:
also `buy_should_work_if_user_has_sufficient_balance_in_pallet_assets` , just need to create a new asset and using pallet_assets and set balance. (refer to pallet_assets tests?)

aang114:
> `create_account` should work TODOs are for the future

Oh I see. Thanks

Oh yeah and previously it was using `reserve()` - that's why it might have been failing

aang114:
> also `buy_should_work_if_user_has_sufficient_balance_in_pallet_assets` , just need to create a new asset and using pallet_assets and set balance. (refer to pallet_assets tests?)

Oh yeah good idea 😅 (https://paritytech.github.io/substrate/master/pallet_assets/pallet/struct.Pallet.html#impl-Pallet%3CT%2C%20I%3E-1)

- Also I realised that when we create a fragment definition - we don't check if the currency field currently exists (https://fragcolor-xyz.github.io/clamor/doc/pallet_fragments/struct.FragmentMetadata.html)!! That's a bug!
    - Also, related to that - shouldn't the currency field be owned by the Fragment Definition's Vault's Account ID?

aang114:
> Yes, I would use the ignore macro on all of them.

Okay, will do that then. Thanks

aang114:
> > also `buy_should_work_if_user_has_sufficient_balance_in_pallet_assets` , just need to create a new asset and using pallet_assets and set balance. (refer to pallet_assets tests?)
> 
> Oh yeah good idea 😅 (https://paritytech.github.io/substrate/master/pallet_assets/pallet/struct.Pallet.html#impl-Pallet%3CT%2C%20I%3E-1)
> 
> * Also I realised that when we create a fragment definition - we don't check if the currency field currently exists (https://fragcolor-xyz.github.io/clamor/doc/pallet_fragments/struct.FragmentMetadata.html)!! That's a bug!
>   
>   * Also, related to that - shouldn't the currency field be owned by the Fragment Definition's Vault's Account ID?

@sinkingsugar

sinkingsugar:
> > also `buy_should_work_if_user_has_sufficient_balance_in_pallet_assets` , just need to create a new asset and using pallet_assets and set balance. (refer to pallet_assets tests?)
> 
> Oh yeah good idea 😅 (https://paritytech.github.io/substrate/master/pallet_assets/pallet/struct.Pallet.html#impl-Pallet%3CT%2C%20I%3E-1)
> 
> * Also I realised that when we create a fragment definition - we don't check if the currency field currently exists (https://fragcolor-xyz.github.io/clamor/doc/pallet_fragments/struct.FragmentMetadata.html)!! That's a bug!
>   
>   * Also, related to that - shouldn't the currency field be owned by the Fragment Definition's Vault's Account ID?

No? Why so?

alebaffa (approved):

alebaffa (changes requested):
Please take care of the last comments by Gio.

pallets/fragments/src/tests.rs
```diff
@@ -1,127 +1,1900 @@
...
+			let mut buy = dd.buy_non_unique;
+
+			buy.publish.definition.metadata.currency = Some(69);
+
+			assert_ok!(upload(dd.account_id, &buy.publish.definition.proto_fragment));
+			assert_ok!(create(dd.account_id, &buy.publish.definition));
+			assert_ok!(publish_(dd.account_id, &buy.publish));
+
+			// Deposit `quantity` to buyer's account
+			todo!();
```
sinkingsugar: still todo?

alebaffa (approved):
👍🏻

PR #126, opened 2022-07-28
Get protos by trait info

alebaffa:
This PR increases the granularity of `get_protos` in case the category in input is `Categories::Trait(ShardsTraitInfo)`.
It makes it possible to search Protos by `ShardsTraitInfo.name` or `ShardsTraitInfo.id` using these patterns:
- if the `Category::Trait` in input contains `ShardsTraitInfo.id = [0u8; 16]`, then `get_protos` will search by `ShardsTraitInfo.name`
- if the `Category::Trait` in input contains `ShardsTraitInfo.name = ""`, then it will search by `ShardsTraitInfo.id`

### DIFF
--- a/pallets/fragments/src/lib.rs
+++ b/pallets/fragments/src/lib.rs
@@ -442,8 +442,12 @@ pub mod pallet {
 			// create vault account
 			// we need an existential amount deposit to be able to create the vault account
 			let vault = Self::get_vault_id(hash);
-			let min_balance = <T as pallet_assets::Config>::Currency::minimum_balance();
-			<T as pallet_assets::Config>::Currency::reserve(&vault, min_balance)?;
+			let min_balance =
+				<pallet_balances::Pallet<T> as Currency<T::AccountId>>::minimum_balance();
+			let _ = <pallet_balances::Pallet<T> as Currency<T::AccountId>>::deposit_creating(
+				&vault,
+				min_balance,
+			);
 
 			let fragment_data = FragmentDefinition {
 				proto_hash,
@@ -955,10 +959,10 @@ pub mod pallet {
 			let frag_account = Self::get_fragment_account_id(class, edition, copy);
 			let min_balance =
 				<pallet_balances::Pallet<T> as Currency<T::AccountId>>::minimum_balance();
-			<pallet_balances::Pallet<T> as ReservableCurrency<T::AccountId>>::reserve(
+			let _ = <pallet_balances::Pallet<T> as Currency<T::AccountId>>::deposit_creating(
 				&frag_account,
 				min_balance,
-			)?;
+			);
 
 			// TODO Make owner pay for deposit actually!
 			// TODO setup proxy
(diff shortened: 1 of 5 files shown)


alebaffa:
Javascript tests failing. Need to check them.

alebaffa:
It was a random error.

pallets/fragments/src/lib.rs
```diff
@@ -442,8 +442,12 @@ pub mod pallet {
 			// create vault account
 			// we need an existential amount deposit to be able to create the vault account
 			let vault = Self::get_vault_id(hash);
-			let min_balance = <T as pallet_assets::Config>::Currency::minimum_balance();
-			<T as pallet_assets::Config>::Currency::reserve(&vault, min_balance)?;
+			let min_balance =
```
sinkingsugar, line 445: this is unrelated no?
Also @aang114 fixed it and this would also not work iirc?

pallets/fragments/src/lib.rs
```diff
@@ -955,10 +959,10 @@ pub mod pallet {
 			let frag_account = Self::get_fragment_account_id(class, edition, copy);
 			let min_balance =
 				<pallet_balances::Pallet<T> as Currency<T::AccountId>>::minimum_balance();
-			<pallet_balances::Pallet<T> as ReservableCurrency<T::AccountId>>::reserve(
+			let _ = <pallet_balances::Pallet<T> as Currency<T::AccountId>>::deposit_creating(
```
sinkingsugar, line 962: @aang114 double check with you

pallets/protos/src/lib.rs
```diff
@@ -251,6 +251,7 @@ pub mod pallet {
 		StakeLocked,
 		/// Cannot find FRAG link to use as stake funds
 		NoFragLink,
+		OwnerNotFound,
```
sinkingsugar: hmm we generally use those errors for extrinsics. Maybe we need a new struct?

sinkingsugar (changes requested):
LGTM

Just I noticed some unrelated code etc

pallets/fragments/src/lib.rs
alebaffa: Yes, it's unrelated. I think it came up merging locally my branch with devel.

pallets/fragments/src/lib.rs
alebaffa: Also this one.

pallets/protos/src/lib.rs
alebaffa: Oh, this can be removed actually. I am not using it.

pallets/fragments/src/lib.rs
sinkingsugar: Weird if was merged should not show though 🤔

alebaffa: Weird, agreed. Trying to figure out how to remove these two files from this PR (which are already on devel).

alebaffa:
Let me quickly do another PR and close this one. I don't want to mess things up for just three files...

PR #131, opened 2022-08-01
add Traits storage to Protos

sinkingsugar:
Add a way to fetch all traits

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -30,12 +30,15 @@ pub use pallet::*;
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
 
-use sp_io::{hashing::blake2_256, transaction_index};
+use sp_io::{
+	hashing::{blake2_128, blake2_256},
+	transaction_index,
+};
 use sp_std::{collections::btree_map::BTreeMap, vec, vec::Vec};
 
 pub use weights::WeightInfo;
 
-use sp_clamor::{get_locked_frag_account, Hash256};
+use sp_clamor::{get_locked_frag_account, Hash128, Hash256};
 
 use scale_info::prelude::{
 	format,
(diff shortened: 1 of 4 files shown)


alebaffa:
This will be useful for filtering protos by Trait too. I will refactor after we merge this one.
👍🏻

pallets/protos/src/lib.rs
```diff
@@ -821,79 +838,61 @@ pub mod pallet {
 			if let Some(struct_proto) = <Protos<T>>::get(proto_id) {
 				if let Some(avail) = avail {
 					if avail && struct_proto.include_cost.is_none() {
-						return false
+						return false;
```
sinkingsugar: @alebaffa you want a semicolon here (I added in merge)

pallets/protos/src/lib.rs
```diff
@@ -821,79 +838,61 @@ pub mod pallet {
...
 					}
 				}
 
 				if categories.len() == 0 {
-					return Self::filter_tags(tags, &struct_proto)
-				} else {					
-					return Self::filter_category(tags, &struct_proto, categories)
+					return Self::filter_tags(tags, &struct_proto);
```
sinkingsugar: @alebaffa also here, if you use `return` use semicolon

pallets/protos/src/lib.rs
```diff
@@ -821,79 +838,61 @@ pub mod pallet {
...
+				} else {
+					return Self::filter_category(tags, &struct_proto, categories);
```
sinkingsugar: also here, trick is you could have removed `return` and in that case no need for semicolon

alebaffa: Yeah, right! Sorry for this!

alebaffa (approved):
LGTM 👍🏻

Issue #128, opened 2022-07-31
Filter protos by shards info


alebaffa:
Continuation of [issue #83](https://github.com/fragcolor-xyz/clamor/issues/83).

As a user of the marketplace, 
I want to filter protos by `ShardsScriptInfo`,
so that I can filter by format, requiring shards or implementing shards.

```rust
pub struct ShardsScriptInfo {
    pub format: ShardsFormat,
    pub requiring: Vec<ShardsTrait>,
    pub implementing: Vec<ShardsTrait>,
}
```

PR #132, opened 2022-08-03
Get protos by shards scripts filter (issue #128)

alebaffa:
This makes it possible to get protos by filtering on `ShardsScriptInfo`. 
It matches either by all fields - as it was implemented before -, or by any single field. Regarding the fields `implementing` and `requiring`, it also match with a subset of items contained in the vectors. This allows users to search protos by subsets of shards scripts.

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -843,9 +843,9 @@ pub mod pallet {
 					Categories::Trait(proto_shard_trait) => {
 						if let Categories::Trait(struct_proto_shard_trait) = &struct_proto.category {
 							// search by trait name (the ID must be all zeros)
-							if (proto_shard_trait.id == [0u8; 16] && proto_shard_trait.name == struct_proto_shard_trait.name) || 
+							if (proto_shard_trait.name == struct_proto_shard_trait.name) || 
 								// search by trait id (the name must be empty)
-								(proto_shard_trait.name.is_empty() && proto_shard_trait.id == struct_proto_shard_trait.id) || 
+								(proto_shard_trait.id == struct_proto_shard_trait.id) || 
 								// search by full match of trait info
 								(proto_shard_trait == struct_proto_shard_trait)
 							{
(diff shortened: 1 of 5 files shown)


alebaffa:
Actually, I am wondering if this matching approach could also be applied to the filter by `Categories::Trait`. Instead of passing an empty `Trait.name` to search by `Trait.ID`, or a `Trait.ID = [0u8; 16]` to search by `Trait.name`, I could just match by names or IDs in case one of them is equal to the Trait given in input.

pallets/protos/src/lib.rs
```diff
@@ -859,6 +859,24 @@ pub mod pallet {
...
+							if !implementing_diffs.is_empty() || !requiring_diffs.is_empty() || (param_script_info.format == stored_script_info.format) || (param_script_info == stored_script_info)
+							{
+								return Self::filter_tags(tags, struct_proto)
+							}
+							else {
+								return false
+							}
+						} else {
+							// it should never go here
+							return false
```
aang114, line 877: shouldn't we make `filter_category()` return a `Result<bool, _>` then instead of `bool`, for better readability and for consistency?

aang114: and for correctness

alebaffa: The function is only internal and it is called by `filter_proto`, which also returns a `bool `. So, if we change `filter_category` we would need to also change `filter_proto`.
I would leave it as it is now.

sinkingsugar (approved):
LGTM, let's try it this way and see how it works in the field

alebaffa:
Changed the logic of filtering by Category Trait or ShardsScriptInfo. Now the filtering will match either the name or the ID, or both, regardless of the value of the other fields. So, no need anymore to send an empty Trait name to search by ID, or a Id of zeroes to search by Trait name.
Same for search by Category ShardsScriptInfo. If one of its values (or all values) match with one Category of a stored ProtoFragment, it will be returned in output.

PR #133, opened 2022-08-05
Traits review

sinkingsugar:
Review traits and improve storage

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -20,7 +20,7 @@ mod tests;
 
 mod weights;
 
-use protos::categories::Categories;
+use protos::{categories::Categories, traits::Trait};
 
 use sp_core::{ecdsa, H160, U256};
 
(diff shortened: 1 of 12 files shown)


pallets/protos/src/lib.rs
```diff
@@ -845,25 +867,6 @@ pub mod pallet {
 			let found: Vec<_> = categories
 				.into_iter()
 				.filter(|cat| match cat {
-					Categories::Trait(proto_shard_trait) => {
```
alebaffa, line 848: Though, this removes completely the filter by trait name or id which is a feature that I understood was needed, no?

alebaffa: Also mentioned here https://github.com/fragcolor-xyz/asset-store/issues/37

alebaffa: So maybe we could leave that .filter by Categories::Trait and use your new structure `<Traits<T>>::contains_key(&trait_id)`

sinkingsugar: but name is not available anymore, you can pass ID still no?

sinkingsugar: basically now we would pass right IDs

alebaffa: Oh yes I see, just by matching the objects will work now since there is only one field. Ok!

alebaffa (approved):

sinkingsugar:
Thank you @alebaffa 🙏👍👍👍👍

PR #134, opened 2022-08-06
use XX64 for Trait ID

sinkingsugar:


alebaffa (approved):
LGTM 👍🏻

sinkingsugar:
> LGTM 👍🏻 

So fast haha 🤩🥳
Btw I noticed many repetitions in the tests, which is ok as we are crunching but maybe we could have some stories ready for future newcomers to do such refactor 🙏

alebaffa:
> So fast haha

Taking advantage from family sleeping 😄 

> we could have some stories ready for future newcomers to do such refactor

Yes, definitely! I did not take much care of refactoring tests so far. Good idea!

PR #137, opened 2022-08-08
Ensure the currency exists before creating Fragment Definition

aang114:
Ensure the `currency` field of the `metadata` parameter of `create()` exists in `pallet_assets` (if it is not `None`)


aang114:
@sinkingsugar could you review this. This is a bug I found when writing Pallet Fragment tests. 

So I want this to be merged first and then the Pallet Fragment Tests PR. Thanks

sinkingsugar (approved):

PR #146, opened 2022-08-28
(Suggestion) Throw Warnings as Errors

aang114:


alebaffa:
Is this after a discussion with Gio?

aang114:
> Is this after a discussion with Gio?

Yeah. But he's not sure if we should throw warnings as errors. What do you think?

alebaffa:
Some times it happens that some warnings might be necessary to keep. It's not our case now, but I generally prefer to have an internal policy for warnings instead of raising errors in compilation.

aang114:
> Some times it happens that some warnings might be necessary to keep. It's not our case now, but I generally prefer to have an internal policy for warnings instead of raising errors in compilation.

Okay sure, makes sense. So should close this PR (without merging) @sinkingsugar ?

PR #147, opened 2022-08-28
CI Throws Error If Any Public Function/Struct/Enum/Module is Not Documented

aang114:
Note: Crate-Level Documentation is the documentation added to the first few lines of the `lib.rs` file of the Crate

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Check Build
         run: |
-          cargo check
+          RUSTFLAGS="-D missing-docs" cargo check
 
       - name: Check Build for Benchmarking
         run: |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4201,6 +4201,7 @@ dependencies = [
  "frame-benchmarking",
  "frame-support",
  "frame-system",
+ "itertools",
  "log",
  "pallet-accounts",
  "pallet-assets",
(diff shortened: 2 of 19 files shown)


alebaffa:
Is this after a discussion?

aang114:
> Is this after a discussion?

I just found this online and thought: "why not?". what do you think 😅 ?

alebaffa:
Well, this could make sense. Even if, same as warnings, I generally prefer to have an internal policy about documentation..but this could make more sense.

alebaffa:
Probably, though, you may want to run this new build locally and see what are the missing docs...and then update the PR with all the missing ones too.

aang114:
This will force us to write documentation for any extrinsic / public function that we add - otherwise the CI fails! So hopefully this reduces any time spent asking each other about a particular function/extrinsic/etc

sinkingsugar:
> This will force us to write documentation for any extrinsic / public function that we add - otherwise the CI fails! So hopefully this reduces any time spent asking each other about a particular function/extrinsic/etc

title is misleading if this is the case.. which is annoying

aang114:
> > This will force us to write documentation for any extrinsic / public function that we add - otherwise the CI fails! So hopefully this reduces any time spent asking each other about a particular function/extrinsic/etc
> 
> title is misleading if this is the case.. which is annoying

Title has been changed

alebaffa (approved):

PR #148, opened 2022-08-31
Custom Metadata for Definitions + RPC Calls in Pallet-Fragments

aang114:


aang114:
Hi @sinkingsugar , please let me know if this is what you had in mind for `getDefinitions` and `getInstances` (that'll be used by the website). **`getDefinitions` might be particularly interesting for you to observe - because I'm using `getProtos` indirectly.** 

Also please let me know if this is what you had in mind for the `custom_metadata` field of `FragmentDefinition`. **I feel we should rename the `metadata` field (and the `FragmentMetadata` struct) to something else**

pallets/fragments/src/lib.rs
```diff
@@ -102,6 +136,8 @@ pub struct FragmentDefinition<TFungibleAsset, TAccountId, TBlockNum> {
 	pub creator: TAccountId,
 	/// The block number when the item was created
 	pub created_at: TBlockNum,
+	/// **Map** that maps the **Key of a Proto-Fragment's Custom Metadata Object** to the **Hash of the aforementioned Custom Metadata Object**
+	pub custom_metadata: BTreeMap<Compact<u64>, Hash256>,
```
sinkingsugar, line 188: OK to be clear, this is ok but it's basically the metadata of the "collection"

pallets/fragments/src/lib.rs
```diff
@@ -474,6 +513,82 @@ pub mod pallet {
...
+		/// Furthermore, this function also indexes `data` in the Blockchain's Database and stores it in the IPFS
+		///
+		/// # Arguments
+		///
+		/// * `origin` - The origin of the extrinsic / dispatchable function
+		/// * `fragment_hash` - **ID of the Fragment Definition**
+		/// * `metadata_key` - The key (of the key-value pair) that is added in the BTreeMap field `custom_metadata` of the existing Fragment Definition's Struct Instance
+		/// * `data` - The hash of `data` is used as the value (of the key-value pair) that is added in the BTreeMap field `custom_metadata` of the existing Fragment Definition's Struct Instance
+		#[pallet::weight(50_000)]
+		pub fn set_custom_metadata(
```
sinkingsugar: I think that this needs to become `set_definition_metadata` as we might want more metadata (tbd)

aang114: true

pallets/fragments/Cargo.toml
```diff
@@ -103,7 +103,7 @@ version = '4.0.0-dev'
 [dependencies]
 sp-clamor = { version = '0.1.0', path = '../../primitives/clamor', default-features = false }
 hex = { version = "0.4.3", default-features = false }
-serde = { version = "1.0.136", features = ["derive"], optional = true }
+serde = { version = "1.0.136", default-features = false, features = ["derive"]} # , optional = true }
```
sinkingsugar: why we need this when no_std?

sinkingsugar: please add conditional #[derive], don't need to do this

aang114: We want the JSON serialization of `FragmentInstances` in the **runtime function** `get_instances()`

So we need serde in a runtime/no_std environment

aang114:
This PR does 4 things:
1. adds a `custom_metadata` field to the `FragmentDefinition` struct
2. adds a `metadata` field to the `FragmentInstance` struct
3. implements the RPC `fragments.getDefinitions()`
4. implements the RPC `fragments.getInstances()`

Sorry for the information overload @alebaffa

pallets/fragments/rpc/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+			only_return_first_copies: params.only_return_first_copies,
+		};
+
+		let result = api.get_instances(&at, params_no_std).map(|list_bytes| {
+			list_bytes.map(|list_bytes| String::from_utf8(list_bytes).unwrap_or(String::from("")))
+		});
+		match result {
+			Err(e) => Err(runtime_error_into_rpc_err(e)),
+			Ok(result) => match result {
+				Err(e) => Err(runtime_error_into_rpc_err(e)),
```
alebaffa, line 127: Can this `Err(e)` happen? Isn't it already handled with the other `Err(e) => Err(...)` in L124?

aang114: lol true

aang114:
(After this PR is merged, I will pur PR #168 up for review - which adds the RPC `protos.getGenealogy()`)

pallets/fragments/src/lib.rs
```diff
@@ -492,6 +593,169 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// **Alters** the **custom metadata** of a **Fragment Definition** (whose ID is `fragment_hash`) by **adding or modifying a key-value pair** (`metadata_key.clone`,`blake2_256(&data.encode())`)
+		/// to the **BTreeMap field `custom_metadata`** of the **existing Fragment Definition's Struct Instance**.
+		/// Furthermore, this function also indexes `data` in the Blockchain's Database and stores it in the IPFS
```
alebaffa, line 597: > and stores it in the IPFS

Does it really do this? Same question for `set_instance_metadata` function.

pallets/fragments/src/lib.rs
```diff
@@ -323,12 +405,30 @@ pub mod pallet {
 	pub type Expirations<T: Config> =
 		StorageMap<_, Twox64Concat, T::BlockNumber, Vec<(Hash128, Compact<Unit>, Compact<Unit>)>>;
 
+	/// **StorageMap** that maps a **Fragment Definition ID and a Number** to a **Data Hash**
```
alebaffa, line 408: I think it is better to specify what "_a Number_" is (a transaction index, correct?). It is not clear here and so the below comment, which advises to see this in order _to understand what "Number" means_, it's not useful.

pallets/fragments/src/lib.rs
```diff
@@ -492,6 +593,169 @@ pub mod pallet {
...
+			// TO REVIEW
+			ensure!(!<DetachedHashes<T>>::contains_key(&proto_hash), Error::<T>::Detached); // Ensure `proto_hash` isn't detached
+
+			let data_hash = blake2_256(&data);
+
+			// we need this to index transactions
+			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
+				.ok_or(Error::<T>::SystematicFailure)?;
+
+			// Write STATE from now, ensure no errors from now...
```
alebaffa, line 637: > // Write STATE from now, ensure no errors from now..

I see this comment a bit everywhere in our codebase. Not clear what it means.

pallets/fragments/src/lib.rs
```diff
@@ -492,6 +593,169 @@ pub mod pallet {
...
+			let proto: Proto<T::AccountId, T::BlockNumber> =
+				<Protos<T>>::get(proto_hash).ok_or(Error::<T>::ProtoNotFound)?;
+			let proto_owner: T::AccountId = match proto.owner { // Get `proto_owner` from `proto`
+				ProtoOwner::User(owner) => Ok(owner),
+				_ => Err(Error::<T>::ProtoOwnerNotFound),
+			}?;
+			ensure!(who == proto_owner, Error::<T>::NoPermission); // Ensure `who` is `proto_owner`
+
+			// TO REVIEW
+			ensure!(!<DetachedHashes<T>>::contains_key(&proto_hash), Error::<T>::Detached); // Ensure `proto_hash` isn't detached
```
alebaffa, line 629: ...and what if the proto hash has already been detached? Could it make sense to trigger the detach function here to update it? Then it would be up to the user to take care of it in the other blockhain.
I think that something needs to happen here, if that is the case. Not sure what, though...

pallets/fragments/src/lib.rs
```diff
@@ -1203,14 +1468,280 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Implementation Block of `Pallet` specifically for RPC-related functions
+	impl<T : Config> Pallet<T>
+		where
+			T::AccountId: UncheckedFrom<T::Hash> + AsRef<[u8]>,
+	{
+
+
+
+		// pub fn get_definitions_old(params: GetDefinitionsParams<T::AccountId, Vec<u8>>) -> Result<Vec<u8>, Vec<u8>> {
```
alebaffa, line 1537: All this commented part can be deleted if it is no more used/necessary. We can get it back from git, in case..

pallets/fragments/src/lib.rs
```diff
@@ -1203,14 +1468,280 @@ impl<T: Config> Pallet<T> {
...
+		//
+		// 	let result = json!(map_definitions).to_string();
+		//
+		// 	Ok(result.into_bytes())
+		// }
+
+		/// **Query** and **Return** **Fragmnent Definition(s)** based on **`params`**
+		pub fn get_definitions(params: GetDefinitionsParams<T::AccountId, Vec<u8>>) -> Result<Vec<u8>, Vec<u8>> {
+
+
```
alebaffa: too many blank lines :)

pallets/fragments/src/lib.rs
```diff
@@ -1203,14 +1468,280 @@ impl<T: Config> Pallet<T> {
...
+					let definition_metadata = definition_struct.custom_metadata;
+					let map_of_matching_metadata_keys = pallet_protos::Pallet::<T>::get_map_of_matching_metadata_keys(&params.metadata_keys, &definition_metadata);
+					(*map_definition).insert("metadata".into(), map_of_matching_metadata_keys.into());
+					// (*map_definition).append(&mut map_of_matching_metadata_keys);
+				}
+
+
+			}
+
+
```
alebaffa: too many blank lines :)
`> cargofmt`

pallets/fragments/src/tests.rs
```diff
@@ -2076,3 +2081,212 @@ mod create_account_tests {
...
+					&FragmentsPallet::get_instances(GetInstancesParams {
+						definition_hash: hex::encode(give.mint.definition.get_definition_id()).into(),
+						limit: u64::MAX,
+						only_return_first_copies: false,
+						..Default::default()
+					}).unwrap()
+				).unwrap(),
+				json!({
+					"1.1": {},
+					"1.2": {},
```
alebaffa, line 2269: How these two copies are found? I cannot see them from the dummydata..

alebaffa (commented):
BTW, the shard scripts are very nice! 👍🏻

pallets/fragments/src/lib.rs
sinkingsugar: Well it better be clear...
Substrate pallets are not safe like smart contracts, everything you write INTO state will stick even if your code MISERABLY FAILS 😄 so you need to ensure no errors, it can re roll transactions but it's not explicit/easy/automatic.

pallets/fragments/src/lib.rs
sinkingsugar: there is no way to determine who is the owner in this case yet.

sinkingsugar (dismissed):
can you always run `cargo fmt` before pushing 🙏🙏🙏

pallets/fragments/src/lib.rs
aang114: Yeah it does. Just like `set_metadata()` in pallet protos

pallets/fragments/src/tests.rs
aang114: When we called `mint()` in this function, one copy was created. (Edition Id 1 , copy ID 1 = "1.1")

When we called `give()` in this function, another copy was created (edition Id 1, copy ID 2 = "1.2")

pallets/fragments/src/lib.rs
aang114: I might need it in the future, if @sinkingsugar wants to filter FDs

alebaffa:
@aang114 please cargofmt and then we are good on this one, I think.

alebaffa:
@aang114 please cargofmt and then we are good on this one, I think.

alebaffa (approved):
@aang114  LGTM! But, please, first cargofmt and then fix the conflict 👍🏻

codecov.yml
```diff
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        threshold: 0.2
+        threshold: 20%
```
sinkingsugar, line 5: What? 20%??

sinkingsugar: We crazy or what?

aang114: Lol. I did it to test whether "0.2" and "20%" are the same thing (cos at the time the CI was failing - so if it passed with "20%" that means they're not the same thing). Clearly they're not the same thing!

I'll change it to 0% soon and only include the "lib.rs" of our custom pallets

PR #153, opened 2022-09-07
Fix filtering get_protos by requiring and implementing

alebaffa:


alebaffa:
Tested also by @paulinechi and it works.

paulinechi (approved):

PR #161, opened 2022-09-08
Remove warnings

alebaffa:
Remove warnings in compilation.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -719,9 +719,11 @@ impl_runtime_apis! {
 			match tx.function {
 				// We want to prevent polluting blocks with a lot of useless invalid data.
 				// TODO perform quick and preliminary data validation
+				#[allow(unused_variables)]
 				Call::Protos(ProtosCall::upload{ref data, ref category, ref tags, ..}) => {
 					// TODO
 				},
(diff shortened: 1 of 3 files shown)


runtime/src/lib.rs
```diff
@@ -719,9 +719,11 @@ impl_runtime_apis! {
 			match tx.function {
 				// We want to prevent polluting blocks with a lot of useless invalid data.
 				// TODO perform quick and preliminary data validation
+				#[allow(unused_variables)]
```
aang114, line 722: instead of adding a macro to ignore the unused variables in runtime

Why don't we just do what the warning says and write it like "tags: _, ..." ?

aang114 (commented):
Why don't we just follow the warning's suggestion in runtime/src/lib.rs?

runtime/src/lib.rs
sinkingsugar: you can't in this case, needs to be right names, this fix is clever actually

sinkingsugar (approved):

runtime/src/lib.rs
alebaffa: Yep, i used the macro on purpose actually..

Issue #174, opened 2022-09-21
Implement periodic vesting for NOVA and Tickets


alebaffa:
Follow the vesting rules described in the whitepaper for both NOVA and Tickets.

Track the block number since the arrival of Lock event from Ethereum and implement a function that allows the account to withdraw/redeem the due amount of Tickets and NOVA gained until that moment.

alebaffa:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/clamor/174?planning-poker) with Zenhub @aang114 @sinkingsugar

alebaffa:
Duplicate of #117

Issue #173, opened 2022-09-21
Implement pallet oracle


alebaffa:
Need an Oracle to get the price of FRAG/USD in order to calculate the correct amount of Tickets and NOVA.

alebaffa:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/clamor/173?planning-poker) with Zenhub @aang114 @sinkingsugar

Issue #145, opened 2022-08-26
Integrate external DEX pallet into Clamor


alebaffa:
Pablo pallet from Composable Finance: https://github.com/ComposableFi/composable/tree/main/frame/pablo provides interesting features that is worth evaluating.

alebaffa:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Whole-Team-623c70cfcdeee40011bd03e3/issues/fragcolor-xyz/clamor/145?planning-poker) with ZenHub @aang114 @sinkingsugar

alebaffa:
`pallet-pablo` has dependencies with
- 7 homemade internal pallets (`composable-maths`, `composable-support`, `composable-test-helpers`, `composable-traits`, `currency-factory`, `fnft`, staking-rewards`)
- 2 forked pallets from FRAME:
    - `pallet-proxy `called `account-proxy` to _enable programmatic access through a standard trait_
    - `pallet-assets` _to provide implementations for common currency traits and functionality for handling transfers and minting. E.g. from [orml](https://docs.rs/orml-traits) or [frame_support](https://github.com/paritytech/substrate/tree/master/frame/support)_

The two forked pallets need to be carefully reviewed. Mostly the `pallet-assets` since we also forked it.

sinkingsugar:
That's a lot of mods... a bit scary.. Can we find a simpler AMM alternative?

alebaffa:
Yeah it seems definitely complicated. Let me see if a simpler one is in my list.

sinkingsugar:
can you check https://hydradx.io/ btw?

alebaffa:
To check this:  [https://github.com/galacticcouncil/HydraDX-node/tree/master/pallets/omnipool ](https://github.com/galacticcouncil/HydraDX-node/tree/master/pallets/omnipool )

PR #175, opened 2022-09-21
Tickets and NOVA: mint, reserve and withdraw.

alebaffa:
### Notes
@guusw @Kryptos-FR @aang114 @sinkingsugar 

- it does not contain a real connection to an Oracle to get the FRAG price yet. So, a hardcoded function returns 1 FRAG = 1 USD. Pallet Oracle will be done in issue #173 
- It also implements issue #174 

## Overview
A user can lock an amount of FRAG tokens on Ethereum mainnet for a certain amout of time (the time is expressed in our [Solidity contract)](https://github.com/fragcolor-xyz/hasten-contracts/blob/clamor/contracts/FragToken.sol):

```solidity
enum Period {
    TwoWeeks,
    OneMonth,
    ThreeMonths,
    SixMonths,
    OneYear
}
``` 

When this is done, our blockchain Clamor, which listens to the related **Lock** event triggered from Ethereum, registers this information in a struct [EthLockedFrag](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L273) and, if the Ethereum account has also an account on our blockchain (we have an internal struct [EVMLinks ](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L295) that stores <Clamor Account, Ethereum Account>), it immediately mints an initial amount of Tickets and NOVA to the related _linked_ account.
 
This initial mint is performed in the function [**internal_lock_update**](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L524), which is a Substrate offchain worker that, for every block, listens to the Ethereum events and perform some operations.

In case there is no such link between Ethereum and Clamor accounts , we reserve some Tickets and NOVA into two storage maps, [EthReservedTickets](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L282) that stores <Ethereum Account, Tickets owned> and [EthReservedNova](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L287) that stores <Ethereum Account, NOVA owned>. These reserved amounts of Tickets and NOVA will be minted to the account immediately when the user decides to create an account on Clamor and to use the [**Link**](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L439) function to link his Ethereum account used to lock FRAG.

Users can lock FRAG whenever they want on Ethereum. Each lock event received on Clamor is handled as separate registered entity.

A user then can decide to **withdraw** all Tickets and NOVA he owns using the [**withdraw** ](https://github.com/fragcolor-xyz/clamor/blob/42d36d3037bff6c649e9817f9ea002309da8f045/pallets/accounts/src/lib.rs#L807)function. While the lock period of FRAG on Ethereum determines how many Tickets and NOVA are yielded on our blockchain and make it not possible to unlock FRAG from Ethereum, the withdraw on our blockchain can happen before or after the lock period (check note below). This way the account can make use of the weekly yielded Tickets and NOVA inside Fragnova network. if they want. 
Consequently, the amount of Tickets and NOVA that is possible to withdraw is relative to the number of weeks that have passed since the FRAG were locked. 
In case there have been multiple locks of FRAG, the weekly amount available to withdraw will be relative to the sum of Tickets and NOVA amounts available depending on the lock period of each lock.
If the withdraw is requested after the lock period is over, the total withdrawable amount will be equal to the max amout that can be yielded within the FRAG lock period.
If the withdraw is requested before the lock period is over, the code registers the week number (in the field [last_withdraw](https://github.com/fragcolor-xyz/clamor/blob/cc4096f582b85a230f714da5e1b1a6b0b9b1b43d/pallets/accounts/src/lib.rs#L167) in the struct [EthLock](https://github.com/fragcolor-xyz/clamor/blob/cc4096f582b85a230f714da5e1b1a6b0b9b1b43d/pallets/accounts/src/lib.rs#L157)) when the last withdraw was performed, and update the information into the struct [here](https://github.com/fragcolor-xyz/clamor/blob/35dca761999b6e427409416e9eb12f1ecc607e53/pallets/accounts/src/lib.rs#L1290). So, the new withdraw will mint only the Tickets and NOVA since the last withdraw.

-----

**A note on withdraw**
Performing a withdraw before the end of the lock period is not a problem for our tokenomics because both Tickets and NOVA are not tradeable (i.e. not available in external crypto exchanges). So even if a user yielded some NOVA and withdraws, he cannot anyway swap them for FRAG since the swap function `NOVA <-> FRAG` can be performed only from our blockchain, and that is where we will perform the check of whether or not the lock period is over.
The swap function is not implemented yet, it will be a work subsequent to issue #145 

### Pallet account
All the logic is contained inside `pallets/account/src/lib.rs`


pallets/accounts/src/lib.rs
```diff
@@ -180,8 +185,11 @@ pub struct AccountInfo<TAccountID, TMoment> {
 pub mod pallet {
 	use super::*;
 	use frame_support::{dispatch::DispatchResult, pallet_prelude::*, Twox64Concat};
+	use frame_support::traits::fungible::Unbalanced;
 	use frame_system::pallet_prelude::*;
 	use sp_runtime::SaturatedConversion;
+	use num_traits::float::FloatCore;
```
sinkingsugar: Floats? For?

pallets/accounts/src/lib.rs
```diff
@@ -404,16 +429,38 @@ pub mod pallet {
 
 			<EVMLinks<T>>::insert(sender.clone(), eth_key);
 			<EVMLinksReverse<T>>::insert(eth_key, sender.clone());
-			let zero: T::Balance = 0u32.saturated_into();
-			<FragUsage<T>>::insert(sender.clone(), zero);
+
+			// Check if the Ethereum account has already some tickets and nova registered when it was not linked
+			if <EthReservedTickets<T>>::contains_key(&eth_key) {
+				// mint tickets
+				let tickets_amount = <EthReservedTickets<T>>::get(&eth_key).unwrap();
```
sinkingsugar: unwrap? you already wrote state above...
Verify first write after

sinkingsugar: Also no unwraps 🙏

pallets/accounts/src/lib.rs
```diff
@@ -404,16 +429,38 @@ pub mod pallet {
...
+					&sender.clone(),
+					tickets_amount)?;
+
+				<EthReservedTickets<T>>::remove(&eth_key);
+				// also emit event
+				Self::deposit_event(TicketsMinted { sender: sender.clone(), balance: tickets_amount });
+			}
+			if <EthReservedNova<T>>::contains_key(&eth_key) {
+				// assign NOVA
+				let nova_amount = <EthReservedNova<T>>::get(&eth_key).unwrap();
```
sinkingsugar: Verify first write after and no unwraps

pallets/accounts/src/lib.rs
```diff
@@ -476,89 +523,102 @@ pub mod pallet {
...
-						// we are good to go, but let's remove the record
-						<EVMLinkVoting<T>>::remove(&data_hash);
-					}
-				} else {
-					// If key `data_hash` doesn't exist in EVMLinkVoting
-					<EVMLinkVoting<T>>::insert(&data_hash, 1);
-					return Ok(());
-				}
-			}
+			let _ = Self::vote_count_and_write(&data_hash, threshold);
```
sinkingsugar: This is a red flag, why result is ignored?

pallets/accounts/src/lib.rs
```diff
@@ -476,89 +523,102 @@ pub mod pallet {
...
+					<pallet_assets::Pallet<T> as Mutate<T::AccountId>>::mint_into(
+						T::TicketsAssetId::get(),
+						&linked,
+						tickets_amount)?;
+					Self::deposit_event(TicketsMinted {sender: linked.clone(), balance: tickets_amount});
+
+					// assign NOVA
+					<pallet_balances::Pallet<T> as Unbalanced<T::AccountId>>::set_balance(
+						&linked,
+						nova_amount)?;
```
sinkingsugar: Verify first write after, tricky cos you have errors above too so you wanna do both ops first

pallets/accounts/src/lib.rs
```diff
@@ -476,89 +523,102 @@ pub mod pallet {
...
+					EthLock { amount: frag_amount, block_number: current_block_number, lock_period },
+				);
 			} else {
 				// If we want to unlock all the FRAG tokens that were
-				// ! TODO TEST
-
 				// if we have any link to this account, then force unlinking
 				let linked = <EVMLinksReverse<T>>::get(sender.clone());
 				if let Some(linked) = linked {
 					Self::unlink_account(linked, sender.clone())?; // Unlink Ethereum Account `sender` and Clamor Account `linked`
```
sinkingsugar: no errors

sinkingsugar: handle the error situation if unavoidable, don't fail the tx

pallets/accounts/src/lib.rs
```diff
@@ -596,7 +656,7 @@ pub mod pallet {
 			// ! Writing state
 
 			let deposit = T::ProxyDepositBase::get() + T::ProxyDepositFactor::get();
-			T::Currency::reserve(&who, deposit)?;
+			<T as pallet_proxy::Config>::Currency::reserve(&who, deposit)?;
```
sinkingsugar, line 823: Notice this error here if fails the transaction won't compromise the state

pallets/accounts/src/lib.rs
```diff
@@ -758,6 +818,15 @@ pub mod pallet {
 			}
 		}
 
+		fn apply_20_percent(amount: u128) -> u128 {
```
sinkingsugar: 20% should not be hardcoded.

pallets/accounts/src/lib.rs
```diff
@@ -881,16 +950,16 @@ pub mod pallet {
 				let amount = data[1].clone().into_uint().ok_or_else(|| "Invalid data")?; // Amount of FRAG token locked/unlocked (`data[1]`)
 
 				// Lock period (`data[2]`). In case of Unlock event, it is zero.
-				let locktime = data[2].clone().into_uint().unwrap_or(U256::from(0));
+				let lock_period = data[2].clone().into_uint().unwrap_or(U256::from(999));
```
sinkingsugar: WHat is 999?

pallets/accounts/src/lib.rs
```diff
@@ -987,5 +1054,28 @@ pub mod pallet {
...
+						// we are good to go, but let's remove the record
+						<EVMLinkVoting<T>>::remove(&data_hash);
+					}
+				} else {
+					// If key `data_hash` doesn't exist in EVMLinkVoting
+					<EVMLinkVoting<T>>::insert(&data_hash, 1);
+					return Ok(());
+				}
+			}
+			Ok(())
```
sinkingsugar: I see all code paths returning Ok()

pallets/accounts/src/lib.rs
alebaffa: Oh, right. I will change it to `Self::vote_count_and_write(&data_hash, threshold)?;` .

pallets/accounts/src/lib.rs
alebaffa: This function is temporary, I need to implement a better way to retrieve all the vesting rules with another PR (issue #174 ).
The float (in another comment above) is needed here indeed.

pallets/accounts/src/lib.rs
alebaffa: Right. I am moving these lines after the if-statements:

```rust
<EVMLinks<T>>::insert(sender.clone(), eth_key);
<EVMLinksReverse<T>>::insert(eth_key, sender.clone());
```

pallets/accounts/src/lib.rs
alebaffa: Since now our smart contract on Ethereum returns the integer representing the enum of the lock periods (`0` = 2 weeks, `1` = 1 month, etc...), I am now setting a default value of `999` in case of Unlock event, where nothing is returned, but a value for this field is still needed later to populate `EthLockUpdate` in `send_unsigned_transaction`. Changing that part is a bit tricky, so I found this solution easier and not dangerous.

This `999` value will not be taken into account in the `internal_lock_event` function where there are all the calculations for internal Frag, Tickets and NOVA.

pallets/accounts/src/lib.rs
alebaffa: Removed. It was just to calculate the percentage, but it's actually useless.

sinkingsugar:
ping me when green and ready to be reviewed.

pallets/accounts/src/lib.rs
```diff
@@ -446,16 +470,13 @@ pub mod pallet {
...
+					Self::deposit_event(TicketsMinted { sender: sender.clone(), balance: amount });
 				}
 			}
 			if <EthReservedNova<T>>::contains_key(&eth_key) {
 				let nova_amount = <EthReservedNova<T>>::get(&eth_key);
 				if let Some(amount) = nova_amount {
-					ensure!(!amount.is_zero(), Error::<T>::AccountAlreadyLinked);
+					ensure!(!amount.is_zero(), Error::<T>::SystematicFailure);
 					// Assign NOVA
 					<pallet_balances::Pallet<T> as Unbalanced<T::AccountId>>::set_balance(
```
sinkingsugar: isn't set_balance dangerous? it "SETS" the balance to the number, shouldn't it be a mint call?

pallets/accounts/src/lib.rs
```diff
@@ -541,7 +559,7 @@ pub mod pallet {
 			let pub_key = &pub_key[12..];
 			ensure!(pub_key == sender.0, Error::<T>::VerificationFailed);
 
-			let data_amount: u128 =
+			let data_amount: u64 =
```
sinkingsugar: hmm why 64 bits?

pallets/accounts/src/lib.rs
```diff
@@ -573,22 +589,46 @@ pub mod pallet {
...
+				initial_nova_amount.saturated_into();
+			let tickets_amount: <T as pallet_assets::Config>::Balance =
+				initial_tickets_amount.saturated_into();
 
-			// TODO - apply mechanism of periodic vesting of NOVA and Tickets
-			let nova_amount: <T as pallet_balances::Config>::Balance = percentage_amount.saturated_into();
-			let tickets_amount: <T as pallet_assets::Config>::Balance = percentage_amount.saturated_into();
+			// need to track when we received the first Lock event from the user.
+			// This is needed in the case the user locks some FRAG, unlock and then Lock again later.
+			let mut first_lock_block_num = current_block_number;
```
sinkingsugar: not sure about this, how can they do that? locks last a definite amount of time

pallets/accounts/src/lib.rs
```diff
@@ -600,7 +640,6 @@ pub mod pallet {
 						&linked,
 						tickets_amount, // amount already ensured to be > 0 in case of Lock
 					)?;
-
 					// assign NOVA
 					<pallet_balances::Pallet<T> as Unbalanced<T::AccountId>>::set_balance(
```
sinkingsugar: not sure again on set_balance

pallets/accounts/src/lib.rs
```diff
@@ -1090,12 +1122,147 @@ pub mod pallet {
...
+							* SECONDS_IN_BLOCK / SECONDS_IN_WEEK +
+						1; // converted to weeks
+					let tickets_amount_to_withdraw =
+						tickets_amount_per_week_at_current_price * num_weeks_since_lock_frag;
+					let nova_amount_to_withdraw =
+						nova_amount_at_current_price * num_weeks_since_lock_frag;
+
+					sp_std::if_std! {
+						// This code is only being compiled and executed when the `std` feature is enabled.
+						println!("frag_lock_block_num: {}", frag_lock_block_number);
```
sinkingsugar: uh? why not use `log::trace!` works in whatever mode

pallets/accounts/src/lib.rs
alebaffa: I think you are looking at the the previous commit. I am not using set balance anymore, i use Balanced deposit(). Mint is not possible with pallet balances, only with assets.

pallets/accounts/src/lib.rs
alebaffa: Because i don't think that the amount of FRAG locked received from Ethereum will ever exceed 64 bit.

pallets/accounts/src/lib.rs
alebaffa: I am thinking about the case of users who lock some FRAG, unlock once the lock period is over and then lock other FRAG again, but never actually link their account on Clamor. 
Probably overthinking....let me delete this.

pallets/accounts/src/lib.rs
alebaffa: I wanted to see logs in the console while testing, i will remove all this once i reopen this PR, i never intended to leave this in the final code.

pallets/accounts/src/lib.rs
alebaffa: I now use deposit() as above.

pallets/accounts/src/lib.rs
alebaffa: ah @sinkingsugar I forgot also to mention that I introduced this `first_lock_block_num` because we use the same structure `EthLockedFrag` for both Lock and Unlock events. So, when the user decide to withdraw (which can happen after or before the `lock_period` is over - if it happens before, the account can withdraw only the amount of Tickets and NOVA yielded so far), without this field, we would not be able to know the block number when the lock of FRAG happened, and so not able to calculate how much to withdraw.

pallets/accounts/src/lib.rs
sinkingsugar: we don't know, FRAG has 18 decimals, make it 128 at least please.

pallets/accounts/src/lib.rs
sinkingsugar: What's deposit?
Shouldn't it be `mint_into`?

pallets/accounts/src/lib.rs
sinkingsugar: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1102
Are we sure?

sinkingsugar: Cos deposit is more internel

pallets/accounts/src/lib.rs
sinkingsugar: good point!

pallets/accounts/src/lib.rs
alebaffa: Done, used mint_into from fungible::Mutate.

pallets/accounts/src/lib.rs
alebaffa: used mint_into yes. Thanks!

pallets/accounts/src/lib.rs
alebaffa: Done!

pallets/accounts/src/dummy_data.rs
```diff
@@ -14,7 +14,35 @@ use sp_core::{
...
 	use frame_support::traits::TypedGet;
-	<Test as crate::Config>::EthChainId::get()
+	<Test as Config>::EthChainId::get()
+}
+
+#[cfg(test)]
+pub fn get_ticket_asset_id() -> u32 {
+	use crate::mock::Test;
+	use frame_support::traits::TypedGet;
+	<Test as Config>::TicketsAssetId::get()
```
Kryptos-FR, line 26: What does it test? I only see a single call.

Shouldn't there be some `assert` with expected values? As well as failure cases.

alebaffa: This is a utility to retrieve the TicketsAssetId, which is defined into the runtime Config
```rust
parameter_types! {
	pub const TicketsAssetId: u64 = 1337;
}
```
It does not test anything, but yeah good idea I should assert the returned value.

alebaffa: Done I added tests for all this config.

pallets/accounts/src/dummy_data.rs
```diff
@@ -14,7 +14,45 @@ use sp_core::{
 fn get_ethereum_chain_id() -> u64 {
 	use crate::mock::Test;
 	use frame_support::traits::TypedGet;
-	<Test as crate::Config>::EthChainId::get()
+	let eth_chain_id = <Test as Config>::EthChainId::get();
+	assert_eq!(eth_chain_id, 5);
```
sinkingsugar: do we care anyway? just curious. As someone might change the parameter for whatever reason and then we fail tests.

pallets/accounts/src/lib.rs
```diff
@@ -159,6 +159,12 @@ pub struct EthLock<TBalance, TBlockNum> {
 	pub amount: TBalance,
 	/// Clamor Block Number in which the locked FRAG tokens was "effectively transfered" to the Clamor Blockchain
 	pub block_number: TBlockNum,
+	/// The block number when FRAG were locked the first time
+	pub first_lock_block_number: TBlockNum,
```
sinkingsugar: isn't `block_number` the same 😅 it should be.

pallets/accounts/src/lib.rs
```diff
@@ -140,8 +140,8 @@ pub struct EthLockUpdate<TPublic> {
 	/// Otherwise, if the event was `Unlock`, it must equal the ***total amount* of FRAG token that was previously locked** on the **FRAG Token Smart Contract**
 	pub amount: U256,
 	/// If the event was `Lock`, it represents the lock period of the FRAG token.
-	/// Owtherwise, if the event was `Unlock`, it is zero.
-	pub locktime: U256,
+	/// If the event was `Unlock`, it is 999.
+	pub lock_period: U256,
```
sinkingsugar: why storing in 32 bytes such small number?

pallets/accounts/src/lib.rs
```diff
@@ -507,57 +603,121 @@ pub mod pallet {
...
 
-			// write this later as unlink_account can fail
 			<EthLockedFrag<T>>::insert(
-				// VERY IMPORTANT TO NOTE
 				sender.clone(),
-				EthLock { amount, block_number: current_block_number },
+				EthLock {
+					amount: frag_amount, // amount already ensured to be > 0 for lock, = 0 for unlock
+					block_number: current_block_number,
+					first_lock_block_number: first_lock_block_num,
```
sinkingsugar: superfluous no? the above is the exact same value

pallets/accounts/src/lib.rs
```diff
@@ -986,5 +1151,193 @@ pub mod pallet {
...
+						)?;
+
+					// remove the locked frag information for this account
+					if num_weeks_since_lock_frag == lock_period_in_weeks {
+						<EthLockedFrag<T>>::remove(&eth_account);
+					} else {
+						// Update the week number of the latest withdraw
+						<EthLockedFrag<T>>::mutate(&eth_account, |eth_lock| {
+							let eth_lock = eth_lock.as_mut().unwrap();
+							eth_lock.block_number = <frame_system::Pallet<T>>::block_number();
```
sinkingsugar: but this value should be immutable, is it the case when multiple locks happen actually the issue?

pallets/accounts/src/dummy_data.rs
alebaffa: I put this after a note from Nicolas. I thought it was useful, but yes I agree.
I can remove these tests here.

pallets/accounts/src/lib.rs
alebaffa: no. `block_number` is updated also with `unlock` event. So, in case the user **lock --> unlock** (both on Ethereum) and then he `withdraw` on Clamor, we would have lost the block number when the lock happened and so not able to mint the correct amount of Tickets and NOVA.
Is this reasoning correct?

pallets/accounts/src/lib.rs
alebaffa: not the same. Explained before. Both Lock and Unlock update the `block_number` field.

pallets/accounts/src/lib.rs
alebaffa: when multiple withdraw happen. A user can decide to withdraw on week 2 and then on week 3. So, in the first withdraw the amount minted is from week 0 --> week 2. In the second withdraw from week 2 --> 3.

alebaffa: Multiple lock is not considered at the moment.

pallets/accounts/src/lib.rs
alebaffa: `U(256)` is 256-bit unsigned integer. [here](https://github.com/paritytech/parity-common/blob/baebea7c3c0f90d825b7024a5ffe97870f819317/primitive-types/src/lib.rs#L38).
But yes I can even use a `u8` here since the range is `[0 - 4]`.

pallets/accounts/src/lib.rs
alebaffa: Implemented.

pallets/accounts/src/lib.rs
```diff
@@ -159,10 +160,8 @@ pub struct EthLock<TBalance, TBlockNum> {
 	pub amount: TBalance,
 	/// Clamor Block Number in which the locked FRAG tokens was "effectively transfered" to the Clamor Blockchain
 	pub block_number: TBlockNum,
-	/// The block number when FRAG were locked the first time
-	pub first_lock_block_number: TBlockNum,
 	/// The FRAG lock period chosen by the user on Ethereum and received from the Lock event
-	pub lock_period: U256,
+	pub lock_period: U128,
```
sinkingsugar: isn't this still too big? Maybe I'm wrong, but isn't this the enum?

sinkingsugar (changes requested):
https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L495
https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L476
I see a lot of possible state corruption, we need to verify first, write later

pallets/accounts/src/lib.rs
alebaffa: Yes it's the enum, but I have issues here with the encoding done by the `ethabi` library to regenerate the Ethereum message. Specifically this part was too difficult to make it right using types other than U256 and U128
```
ethabi::encode(
	&[
		Token::Bytes(lock.data.signature.0.to_vec()),
		Token::Uint(lock.data.amount),
		Token::Uint(U256::from(lock.data.lock_period))
	]
),
```

sinkingsugar: Signature is fine, no need to store it u256 or 128 though.

alebaffa:
> https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L495
> 
> 
> https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L476
> 
> 
> I see a lot of possible state corruption, we need to verify first, write later

All the writes are done at the very end after a lot of checks.
I spent really a huge amount of time trying to _check first, write later_. I think I need some suggestions on how to fix this.

pallets/accounts/src/lib.rs
alebaffa: I changed it to `u8` (as it is sent from Ethereum).

alebaffa:
> https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L495
> 
> 
> https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L476
> 
> 
> I see a lot of possible state corruption, we need to verify first, write later

@sinkingsugar I think I've fixed it.

pallets/accounts/src/lib.rs
alebaffa: I now split the Lock from the Unlock, creating a new type [EthUnlockedFrag](https://github.com/fragcolor-xyz/clamor/blob/6fe5df92349f589002381c55d56f39a6e90ec426/pallets/accounts/src/lib.rs#L289). Having both using the same storage was complicating things in the withdraw calculations.

pallets/accounts/src/benchmarking.rs
```diff
@@ -165,8 +154,8 @@ benchmarks! {
 									&b"FragLock"[..],
 									&get_ethereum_public_key(&ethereum_secret_key_struct).0[..],
 									&T::EthChainId::get().to_be_bytes(),
-									&Into::<[u8; 32]>::into(U256::from(7u32)), // same as `data.amount`
-									&Into::<[u8; 32]>::into(U256::from(7u32)) // same as `data.locktime`
+									&Into::<[u8; 32]>::into(U256::from(100u32)), // same as `data.amount`
```
guusw, line 205: If it's the same might as well make it a constant and just reference it in both places

pallets/accounts/src/benchmarking.rs
```diff
@@ -241,5 +264,70 @@ benchmarks! {
 		assert_eq!(ExternalAuthorities::<T>::get(), BTreeSet::new());
 	}
 
+	// TODO
```
guusw: I'd avoid commented out code, if it's untested or going to be added later: keep it in a branch

pallets/accounts/src/lib.rs
```diff
@@ -141,8 +142,8 @@ pub struct EthLockUpdate<TPublic> {
 	/// Otherwise, if the event was `Unlock`, it must equal the ***total amount* of FRAG token that was previously locked** on the **FRAG Token Smart Contract**
 	pub amount: U256,
 	/// If the event was `Lock`, it represents the lock period of the FRAG token.
-	/// Owtherwise, if the event was `Unlock`, it is zero.
-	pub locktime: U256,
+	/// If the event was `Unlock`, it is 999.
+	pub lock_period: u8,
```
guusw, line 146: 999 can not be represented by an u8 (max=255)
Is this magic value actually checked to see if it was an unlock, since there's also `lock: bool`. maybe just note `This value is unused if the event is an unlock`

pallets/accounts/src/lib.rs
```diff
@@ -407,18 +474,70 @@ pub mod pallet {
 			ensure!(!<EVMLinks<T>>::contains_key(&sender), Error::<T>::AccountAlreadyLinked);
 			ensure!(!<EVMLinksReverse<T>>::contains_key(eth_key), Error::<T>::AccountAlreadyLinked);
 
+			if <EthReservedTickets<T>>::contains_key(&eth_key) {
```
guusw, line 518: Why is it possible to get tickets/nova before having a clamor account?
Why not just require one before being able to lock?

pallets/accounts/src/lib.rs
```diff
@@ -407,18 +474,70 @@ pub mod pallet {
...
+					);
+					ensure!(
+						!nova_old_balance.checked_add(&amount).is_none(),
+						Error::<T>::SystematicFailure
+					);
+				}
+			}
+
+			// Check if the Ethereum account has already some tickets and nova registered when it was not linked
+			if <EthReservedTickets<T>>::contains_key(&eth_key) {
```
guusw, line 543: Can these duplicate if branches be merged?
It doesn't look like they have any specific ordering dependency

pallets/accounts/src/lib.rs
```diff
@@ -993,5 +1169,201 @@ pub mod pallet {
...
+			 // TODO implement pallet Oracle
+		}
+
+		/// Convert the lock period integer retrieved from Ethereum event into the number of weeks.
+		/// Refer to https://github.com/fragcolor-xyz/hasten-contracts/blob/clamor/contracts/FragToken.sol
+		pub fn eth_lock_period_to_weeks(lock_period: u8) -> Result<u8, Error<T>> {
+			let sec = match lock_period {
+				0 => 2,  // 2 weeksu64
+				1 => 4,  // 1 month
+				2 => 12, // 3 months
```
guusw: Probably doesn't matter too much but 3/6 months is 13/26 weeks respectively (would have to change the contract)
Feel free to leave as is

pallets/protos/src/lib.rs
```diff
@@ -797,7 +797,7 @@ pub mod pallet {
 
 			// make sure the user has enough tickets
 			let balance = <pallet_assets::Pallet<T> as Inspect<T::AccountId>>::balance(
-				T::TicketsAssetId::get(),
+				<T as pallet::Config>::TicketsAssetId::get(),
```
guusw, line 828: Does `<T as pallet::Config>::...` change anything compared to `T::...` or does it not compile without?

runtime/src/lib.rs
```diff
@@ -431,6 +431,10 @@ impl pallet_accounts::EthFragContract for Runtime {
 	}
 }
 
+parameter_types! {
+	pub const TicketsAssetId: u64 = 1337;
```
guusw, line 437: This seems duplicated in some places, can they be moved to a common location?

guusw (commented):
Not too familiar with the clamor workings but added some questions and feedback.

pallets/accounts/src/lib.rs
alebaffa: I forgot to remove that comment. It was 999 since its type was bigger (U256 actually). It is not considered in the code.

pallets/accounts/src/lib.rs
alebaffa: The lock is done on Ethereum, not in our Blockchain. So nothing can actually prevent people from locking FRAG using our smart contract there, even if we create a web page to do so. 
For this reason we decided to handle this case too for now.

pallets/accounts/src/benchmarking.rs
alebaffa: Yeah this could actually be extracted from here. I will work on it.

pallets/accounts/src/benchmarking.rs
alebaffa: Right.

pallets/accounts/src/lib.rs
alebaffa: There is no ordering needed in this case. All reserved tokens will be handled the same.

pallets/protos/src/lib.rs
alebaffa: Uh I need to check again why i did this..

runtime/src/lib.rs
alebaffa: This would actually be that common location :) in fact, it is where all the other Config in the other pallets read the value from.
I will check again for duplicates in tests maybe..

pallets/accounts/src/lib.rs
guusw: Yeah but say the ETH contract requires you to specify a clamor address, wouldn't that avoid this case?

pallets/accounts/src/lib.rs
guusw: I'd just merge the if statements then, currently it reads:
```
if <EthReservedTickets<T>>::contains_key(&eth_key) { ... }
if <EthReservedNova<T>>::contains_key(&eth_key) { ...}
if <EthReservedTickets<T>>::contains_key(&eth_key) { ... }
if <EthReservedNova<T>>::contains_key(&eth_key) { ...}
```
change it to
```
if <EthReservedTickets<T>>::contains_key(&eth_key) { ... }
if <EthReservedNova<T>>::contains_key(&eth_key) { ...}
```

runtime/src/lib.rs
guusw: Yeah saw 2 other usages in tests I think, probably reference this value from there.

pallets/accounts/src/lib.rs
alebaffa: What if the Clamor account doesn't exist though..? The caller of the smart contract could put a random address and there wouldn't be any check on Ethereum, creating a weird situation on Clamor to handle.
This is why we prefer to have people using the Link() function to link their Ethereum address to their Clamor ones..and then we do the needed check.

pallets/accounts/src/lib.rs
alebaffa: So, actually I made this on purpose and it is related to how storage in Substrate works. They were merged before a previous review I received (so hopefully this is fine now). In Substrate once the state/storage is written it would not rollback in case the transction fails later. So here the first two if-statements check that there would be no problem to perform the `mint_into()` function calls in the third and fourth if-statements. If they were merged, and the second if-statement fails after the first if-statement was successful, the storage written in the first if-statement would not rollback, causing state incoherence.

pallets/protos/src/lib.rs
alebaffa: this is actually not part of this PR.

runtime/src/lib.rs
alebaffa: I will refactor tests with another PR since it is lower priority and it would require some work that may not be so trivial.

pallets/accounts/src/benchmarking.rs
alebaffa: I will work on this with another PR since benchmarkings are lower priority and require some work.

pallets/accounts/src/lib.rs
guusw: Hmm yeah, was just comparing it to how ETH transfers work for example, It's up to the user to make sure they enter a correct address. (although can still be assisted later by web UI's, etc.)
But yeah the extra complexity might be worth it to avoid user error when entering the wrong address.

pallets/accounts/src/lib.rs
guusw: Yeah then it makes sense to have it this way, thanks for the explanation

pallets/accounts/src/lib.rs
sinkingsugar: good point 👍

sinkingsugar:
> > https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L495
> > 
> > https://github.com/fragcolor-xyz/clamor/blob/26769d9913a47e8fb30f3db417bca6938f6e2b2c/pallets/accounts/src/lib.rs#L476
> > 
> > I see a lot of possible state corruption, we need to verify first, write later
> 
> @sinkingsugar I think I've fixed it.

seems so for what I could see 👍

pallets/accounts/src/lib.rs
alebaffa: Fixing it thanks!

alebaffa:
So, if it's ok i am going to merge this one (after having realigned with devel and squashed all commits).

sinkingsugar (changes requested):
So did we test with real goerli? did you have geth to test and deploy contract?

alebaffa:
@sinkingsugar I deployed on goerli and locked FRAG with my personal Ethereum goerli account and it works. I also found a bug in the Solidity contract that I am going to fix.

alebaffa:
@sinkingsugar If there are no other remarks I am going to squash commits and merge this PR, if you think you can approve it. I will create another dedicated issue to check into the hash generation (and signature validation) in `internal_lock_update` that I still do not manage to figure out and it requires time (using `ethabi.encode` function does not solve it apparently). 
Please, let me know.

sinkingsugar:
I think we need to merge when it's actually correctly working. @alebaffa not sure why it's taking so long, did you change to eip 721?

sinkingsugar:
CI failing, lets have it green before final review

alebaffa:
@sinkingsugar done!

sinkingsugar:
Is it correctly working now? Full circle from eth?

alebaffa:
> Is it correctly working now? Full circle from eth?

@sinkingsugar Now yes!

sinkingsugar (approved):

PR #179, opened 2022-09-22
Added Extrinsic `protos.ban()`

aang114:
This extrinsic deletes the specified Proto-Fragment from all relevant Storage Items

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -458,12 +458,11 @@ pub mod pallet {
 
 		/// Delete Proto-Fragment `proto_hash` from all relevant Storage Items
 		#[pallet::weight(50_000)]
-		pub fn delete_proto(origin: OriginFor<T>, proto_hash: Hash256) -> DispatchResult {
+		pub fn ban(origin: OriginFor<T>, proto_hash: Hash256) -> DispatchResult {
 			ensure_root(origin)?;
 
 			let proto_struct = <Protos<T>>::get(&proto_hash).ok_or(Error::<T>::ProtoNotFound)?;
 
-			<Protos<T>>::remove(&proto_hash);
 			<ProtosByCategory<T>>::mutate(&proto_struct.category, |list_protos| {
 				if let Some(list_protos) = list_protos {
 					list_protos.retain(|current_hash| proto_hash != *current_hash);
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -456,36 +456,6 @@ pub mod pallet {
 		}
 
 
-		/// Delete Proto-Fragment `proto_hash` from all relevant Storage Items
-		#[pallet::weight(50_000)]
-		pub fn ban(origin: OriginFor<T>, proto_hash: Hash256) -> DispatchResult {
-			ensure_root(origin)?;
-
-			let proto_struct = <Protos<T>>::get(&proto_hash).ok_or(Error::<T>::ProtoNotFound)?;
-
-			<ProtosByCategory<T>>::mutate(&proto_struct.category, |list_protos| {
-				if let Some(list_protos) = list_protos {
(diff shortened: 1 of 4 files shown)


pallets/protos/src/lib.rs
```diff
@@ -455,6 +455,37 @@ pub mod pallet {
...
+				}
+			});
+			for account_that_curated_proto  in <ProtoCurations<T>>::iter_key_prefix(proto_hash) {
+				<AccountCurations<T>>::mutate(&account_that_curated_proto, |list_protos| {
+					if let Some(list_protos) = list_protos {
+						list_protos.retain(|current_hash| proto_hash != *current_hash);
+					}
+				});
+			}
+			<ProtoCurations<T>>::clear_prefix(proto_hash, u32::MAX, None);
```
aang114: Full disclosure: I don't know what the last parameter `maybe_cursor` of the function `clear_prefix()` does 😅  @sinkingsugar .

So this might a failure point for this extrinsic

sinkingsugar:
Also remove from Protos main map

pallets/protos/src/lib.rs
sinkingsugar: Hmm let's test this extensively maybe. I'll check tomorrow.

aang114:
> Also remove from Protos main map

lol true. fixed

pallets/protos/src/lib.rs
aang114: hopefully should work since all relevant storage items have been cleared

sinkingsugar: Ah it's simply cos if there is too much to delete it might not fit one block so you spread deletion in many blocks...

sinkingsugar: Should be ok for now like you did

sinkingsugar: Maybe remove_prefix is better btw

aang114: `remove_prefix` is deprecated though? We should still use it?

sinkingsugar: Ah, no then 👍

sinkingsugar:
> > Also remove from Protos main map
> 
> 
> 
> lol true. fixed

Not sure. Might be correct keeping. The intention is to avoid indexing (get_protos rpc visibility)

shards/common.edn
```diff
@@ -9,11 +9,11 @@
 ;; Index of the Upload Function in Protos Pallet
 (def upload-index 0)
 ;; Index of the Patch Function in Protos Pallet
-(def patch-index 1)
+(def patch-index 2)
```
sinkingsugar: No no no, put ban at end. We need to add this as an update as I'm resetting testnet already

sinkingsugar: In general always add new functions at end of pallets cc @alebaffa

sinkingsugar:
Add a test @aang114 so CI will pass, it's missing coverage

sinkingsugar (changes requested):
Add a test @aang114 so CI will pass, it's missing coverage

sinkingsugar (approved):

pallets/protos/src/lib.rs
```diff
@@ -796,6 +797,37 @@ pub mod pallet {
...
+				if let Some(list_protos) = list_protos {
+					list_protos.retain(|current_hash| proto_hash != *current_hash);
+				}
+			});
+			<ProtosByOwner<T>>::mutate(proto_struct.owner, |list_protos| {
+				if let Some(list_protos) = list_protos {
+					list_protos.retain(|current_hash| proto_hash != *current_hash);
+				}
+			});
+			for account_that_curated_proto  in <ProtoCurations<T>>::iter_key_prefix(proto_hash) {
```
sinkingsugar, line 819: Actually I don't think we wanna remove this. It has nothing to do with get_protos anyway and this is strictly for royalties

sinkingsugar (changes requested):

PR #182, opened 2022-09-25
Only Build Clamor Once in the CI

aang114:
This saves time since we are currently building Clamor 2.5 times (2 `cargo build`s and 1 `cargo check`)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -40,7 +40,7 @@ jobs:
           
       - name: Check if Clamor Builds
         run: |
-          RUSTFLAGS="-D missing-docs" cargo build
+          RUSTFLAGS="-D missing-docs" cargo build --features=runtime-benchmarks
           
       - name: Check if all the Polkadot.js RPC tests pass
         run: |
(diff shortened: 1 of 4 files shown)


aang114:
@sinkingsugar is it still too long?

aang114:
> @sinkingsugar is it still too long?

There's only 2 ways I can make it faster now:
1. Running "Push to codecov.io" parallelly
2. combining "Check if Clamor Builds" and "Check Build for Benchmarking" into one step

Lemme know if this makes sense. Thanks

.github/workflows/check.yml
```diff
@@ -37,10 +37,13 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: '14'  # The node-version input is optional. If not supplied, the node version from PATH will be used. However, it is recommended to always specify Node.js version and don't rely on the system one.
-
-      - name: Test RPCs
+          
+      - name: Check if Clamor Builds
+        run: |
+          RUSTFLAGS="-D missing-docs" cargo build
```
alebaffa: why `build` and not `check`?

alebaffa:
IMHO - sorry to jump in - it makes sense to do both, yes.

sinkingsugar:
> IMHO - sorry to jump in - it makes sense to do both, yes.

this

alebaffa (approved):

sinkingsugar:
What's going on here?
Merge "Check if clamor build" with "Check if test pass" with push codecov... why we run ANY code when pushing to codecov. Build clamor with coverage ON from the beginning. Same for tests. So you run coverage at same time.
Push cover should not build and run anything.

sinkingsugar (approved):
GOod enough, let's stop wasting more time.

PR #193, opened 2022-09-30
Fix Benchmarking for all Pallets

aang114:


aang114:
Added benchmarking for all extrinsics in all custom pallets

.github/workflows/check.yml
```diff
@@ -61,10 +61,9 @@ jobs:
         run: |
           # cargo test --no-fail-fast
           cargo install cargo-tarpaulin
-          cargo tarpaulin --out Xml
+          cargo tarpaulin --features runtime-benchmarks --out Xml
```
alebaffa, line 64: `--features runtime-benchmarks` is already in the `Check if Clamor Builds task`, isn't it?
Why do we need to have it here as well?

Cargo.toml
```diff
@@ -10,4 +10,9 @@ members = [
 # chainblocks = { path = "../chainblocks/rust" }
 
 [profile.release]
-panic = 'unwind'
+# Substrate runtime requires unwinding.
+panic = "unwind"
+
+# Inspired from https://github.com/paritytech/substrate/blob/master/Cargo.toml:
+[profile.production]
+inherits = "release"
```
alebaffa, line 18: Is this needed? If it just inherits we can just run `cargo build --profile release`, no?

pallets/accounts/Cargo.toml
```diff
@@ -109,11 +109,11 @@ serde = { version = "1.0.136", features = ["derive"], optional = true }
 serde_json = { version = '1.0.79', default-features = false, features = ['alloc'] }
 ethabi = { version = "17.0.0", default-features = false }
 parking_lot = "0.12.1" # copied from https://github.com/JoshOrndorff/recipes/blob/master/pallets/ocw-demo/Cargo.toml#L11
-libsecp256k1 = {version = "0.7.0", default-features = false} # https://crates.io/crates/libsecp256k1
+libsecp256k1 = {version = "0.7.0", default-features = false} # https://crates.io/crates/libsecp256k1 (libsecp256k1 is owned by Parity Tech by the way)
```
alebaffa, line 112: why was it needed to be specified here?

pallets/accounts/src/mock.rs
```diff
@@ -30,6 +30,13 @@ use sp_runtime::{
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 type Block = frame_system::mocking::MockBlock<Test>;
 
+/// Balance of an account.
+pub type Balance = u128;
+
+pub const MILLICENTS: Balance = 1_000_000_000;
+pub const CENTS: Balance = 1_000 * MILLICENTS; // assume this is worth about a cent.
+pub const DOLLARS: Balance = 100 * CENTS;
```
alebaffa, line 38: Why DOLLARS? It should be NOVA, right?

pallets/accounts/src/weights.rs
```diff
@@ -15,25 +15,27 @@
...
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-02-22, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2022-10-06, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! HOSTNAME: `Karans-Air`, CPU: `<UNKNOWN>`
 //! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("dev"), DB CACHE: 1024
 
 // Executed Command:
-// ./target/release/clamor
+// ./target/production/clamor
```
alebaffa, line 26: Consequence of my comment above about inherits = release in Cargo.toml.

pallets/accounts/src/mock.rs
sinkingsugar: mock.rs, does it matter?

.github/workflows/check.yml
sinkingsugar: hmm I'm more concerned about time, does coverage really improve much? and how about time?

pallets/accounts/src/mock.rs
alebaffa: no of course, just for coherence.

sinkingsugar: then you missed https://github.com/fragcolor-xyz/clamor/blob/devel/runtime/src/lib.rs#L164
I think its more of a convention from substrate devs

.github/workflows/check.yml
aang114: > You can test benchmarks using the same mock runtime that you created for unit testing your pallet. The benchmarking macro you use in your benchmarking.rs module automatically generates test functions for you. (https://docs.substrate.io/test/benchmark/)

By adding `--features runtime-benchmarks`, it tests whether all the benchmarks in all the "benchmarks.rs" files are correct

pallets/accounts/Cargo.toml
aang114: Because there were other `libsecp256k1` rust libraries in "crates.io" and so this is my rationale of why I chose this one 😂

.github/workflows/check.yml
aang114: > hmm I'm more concerned about time, does coverage really improve much? and how about time?

It'll probably increase the time it takes to run by a little bit.

Cargo.toml
aang114: True. I forgot to add a few lines at the bottom like how substrate does it: https://github.com/paritytech/substrate/blob/master/Cargo.toml#L304-L312

Added them now

pallets/accounts/src/weights.rs
aang114: Have answered your previous comment

alebaffa (approved):

Cargo.toml
```diff
@@ -10,4 +10,16 @@ members = [
...
+
+# Sacrifice compile speed for execution speed by using optimization flags:
+
+# https://doc.rust-lang.org/rustc/linker-plugin-lto.html
+lto = "fat"
```
sinkingsugar, line 23: why? makes it very slow to link and we don't need such optimization?

aang114: > The production profile applies various compiler optimizations.
These optimizations slow down the compilation process a lot.
If you are just testing things out and don't need final numbers, use the --release command-line option instead of the production profile. (https://docs.substrate.io/test/benchmark/)

This is only for testing things out

PR #195, opened 2022-10-01
Ensure Proto-Fragment cannot refer itself

### LINKED ISSUE #189: Rename the parameter `new_tags` of `protos.patch()` to `tags`
`new_tags` doesn't make sense, since it actually overwrites the current tags

### LINKED ISSUE #190: Ensure that the `new_references` parameter of `protos.patch()` cannot contain the `proto_hash` parameter
Otherwise, this would cause a circular dependency!

aang114:
See linked issues for description

### DIFF
--- a/pallets/protos/src/lib.rs
+++ b/pallets/protos/src/lib.rs
@@ -317,7 +317,7 @@ pub mod pallet {
 		/// Not enough tokens to stake
 		InsufficientBalance,
 		/// Proto-Fragment's References includes itself!
-		ProtoInReferences,
+		CircularReference,
 	}
 
 	// Dispatchable functions allows users to interact with the pallet and invoke state changes.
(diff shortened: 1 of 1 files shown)


pallets/protos/src/lib.rs
```diff
@@ -316,6 +316,8 @@ pub mod pallet {
 		ReferenceNotFound,
 		/// Not enough tokens to stake
 		InsufficientBalance,
+		/// Proto-Fragment's References includes itself!
+		ProtoInReferences,
```
sinkingsugar: CircularReference

sinkingsugar (changes requested):

pallets/protos/src/lib.rs
aang114: okay. renamed and pushed

sinkingsugar:
~~As it is this PR is pointless.~~
~~The check will never ever be triggered.~~
I get it now. It's simply to prevent putting `upload` hash also as ref.

sinkingsugar (approved):

PR #205, opened 2022-10-10
Validate Extrinsic before it is gossiped or applied

aang114:
This PR does the following 2 things:

- Ensures that the extrinsic does not contain an invalid call or does not exceed maximum depth length **before it is gossiped**. This is done in `<Runtime as TaggedTransactionQueue<Block>>::validate_transaction()`
- Ensures that the extrinsic does not contain an invalid call or does not exceed maximum depth length **before it is applied**. This is done in `<Runtime as sp_block_builder::BlockBuilder<Block>>::apply_extrinsic()`

P.S: This PR also limits the maximum length (in bytes) that the metadata data of a Proto-Fragment / Fragment Definition / Fragment Instance can be


runtime/src/lib.rs
```diff
@@ -752,6 +1003,115 @@ pub type Executive = frame_executive::Executive<
...
+			},
+		},
+		Categories::Shards(shards_script_info_struct) => {
+			let format = shards_script_info_struct.format;
+			let requiring = &shards_script_info_struct.requiring;
+			let implementing = &shards_script_info_struct.implementing;
+
+			let all_required_trait_impls_found = requiring.iter().all(|shards_trait| {
+				proto_references.iter().any(|proto| {
+					if let Some(trait_impls) = pallet_protos::TraitImplsByShard::<Runtime>::get(proto) {
```
aang114: @sinkingsugar - seems it is possible access a pallet's storage item in `validate_transaction()` 🤔. 

(this function `is_valid()` is called by `<Runtime as TaggedTransactionQueue<Block>::validate_transaction()` btw)

runtime/src/lib.rs
```diff
@@ -752,6 +1003,115 @@ pub type Executive = frame_executive::Executive<
...
+					} else {
+						false
+					}
+				})
+			});
+
+			let all_traits_implemented_in_this_shards = implementing.iter().all(|shards_trait| {
+				match format {
+					ShardsFormat::Edn => {
+						// TODO - How do I check if this Shards is implementing this Trait - Giovanni Petrantoni
```
aang114: @sinkingsugar - how exactly do I verify if this shards script is implementing all the traits it specified in `Categories::Shards::implementing` 😅 ? Thanks :)

runtime/src/lib.rs
```diff
@@ -752,6 +1003,115 @@ pub type Executive = frame_executive::Executive<
...
+		Categories::Vector(sub_categories) => match sub_categories {
+			VectorCategories::SvgFile => false, // TODO - Giovanni Petrantoni
+			VectorCategories::TtfFile => infer::is(data, "ttf"), // ttf_parser::Face::parse(&data[..], 0).is_ok(),
+		},
+		Categories::Video(sub_categories) => match sub_categories {
+			VideoCategories::MkvFile => infer::is(data, "mkv"),
+			VideoCategories::Mp4File => infer::is(data, "mp4"),
+		},
+		Categories::Model(sub_categories) => match sub_categories {
+			ModelCategories::GltfFile => false, // TODO - Giovanni Petrantoni
```
aang114: @sinkingsugar - what do we do about all file types whose file signature / magic number (https://en.wikipedia.org/wiki/List_of_file_signatures) **does not exist in the `infer` crate?**

Should I manually verify them (i.e search online for their magic number and verify it against `data`) or just leave them as `false` - since anyways this is literally just a band-aid solution that won't be used for the real thing 😅?

(Btw all the file types that can't be verified using `infer` crate are labelled as `// TODO - Giovanni Petrantoni` 😅 )

runtime/src/lib.rs
```diff
@@ -752,6 +1003,115 @@ pub type Executive = frame_executive::Executive<
...
+			ModelCategories::Sdf => infer::is(data, "sqlite"), // TODO Review - We are not checking for the file extension ".sdf" https://en.wikipedia.org/wiki/Spatial_Data_File
+			ModelCategories::PhysicsCollider => false, // TODO - Giovanni Petrantoni
```
aang114: @sinkingsugar  -  no idea what the file extension for this is even supposed to be 😅

runtime/src/lib.rs
```diff
@@ -752,6 +1003,115 @@ pub type Executive = frame_executive::Executive<
...
+			VectorCategories::SvgFile => false, // TODO - Giovanni Petrantoni
+			VectorCategories::TtfFile => infer::is(data, "ttf"), // ttf_parser::Face::parse(&data[..], 0).is_ok(),
+		},
+		Categories::Video(sub_categories) => match sub_categories {
+			VideoCategories::MkvFile => infer::is(data, "mkv"),
+			VideoCategories::Mp4File => infer::is(data, "mp4"),
+		},
+		Categories::Model(sub_categories) => match sub_categories {
+			ModelCategories::GltfFile => false, // TODO - Giovanni Petrantoni
+			ModelCategories::Sdf => infer::is(data, "sqlite"), // TODO Review - We are not checking for the file extension ".sdf" https://en.wikipedia.org/wiki/Spatial_Data_File
```
aang114: this is only a partial solution btw since we're not checking for the file extension `.sdf`. Is that okay ?

runtime/src/lib.rs
sinkingsugar: Don't need to put my name lol

runtime/src/lib.rs
sinkingsugar: Our format

runtime/src/lib.rs
sinkingsugar: sdf is our format, not sqlite, set false

runtime/src/lib.rs
sinkingsugar: you don't for now it's fine

aang114:
This PR does the following things:
1. Ensures that if the `data: Vec<u8>` param of certain calls is incorrect (see `does_call_have_correct_data_param()` for check logic)
   - an extrinsic with this call is never gossiped (ensured using `validate_transaction()`) 
   - the call is never applied (ensured using `<BaseCallFilter as Contains>::contains()`). 
2. Ensures that an extrinsic with a call that  has a depth level greater than 4 (see  `does_call_exceed_maximum_depth_level()` for check logic)
   -  it is never applied (ensured in `apply_extrinsic()`) and 
   - it is never gossiped (ensured in `validate_transaction()`)
3. Ensures that if a batch call contains uses `transaction_index::index`, it is never applied.

aang114:
Hi @sinkingsugar, please approve this PR (if okay) so that it can be merged. Thanks

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+							false
+						}
+					})
+				});
+
+				let all_traits_implemented_in_this_shards = implementing.iter().all(|_shards_trait| {
+					match format {
+						ShardsFormat::Edn => {
+							// TODO - How do I check if this Shards is implementing this Trait - Giovanni Petrantoni
+							false
```
sinkingsugar: true, let it pass

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+						},
+						ShardsFormat::Binary => {
+							// TODO - How do I check if this Shards is implementing this Trait - Giovanni Petrantoni
+							false
```
sinkingsugar: true, let it pass for now

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+			Categories::Vector(sub_categories) => match sub_categories {
+				VectorCategories::SvgFile => false,
+				VectorCategories::TtfFile => infer::is(data, "ttf"), // ttf_parser::Face::parse(&data[..], 0).is_ok(),
+			},
+			Categories::Video(sub_categories) => match sub_categories {
+				VideoCategories::MkvFile => infer::is(data, "mkv"),
+				VideoCategories::Mp4File => infer::is(data, "mp4"),
+			},
+			Categories::Model(sub_categories) => match sub_categories {
+				ModelCategories::GltfFile => false,
```
sinkingsugar, line 467: `.glb` cannot?

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+				ModelCategories::Sdf => false,
+				ModelCategories::PhysicsCollider => false, // "This is a Fragnova/Fragcolor data type" - Giovanni Petrantoni
+			},
+			Categories::Binary(sub_categories) => match sub_categories {
+				BinaryCategories::WasmProgram => infer::is(data, "wasm"), // wasmparser_nostd::Parser::new(0).parse_all(data).all(|payload| payload.is_ok()), // REVIEW - shouldn't I check if the last `payload` is `Payload::End`?
+				BinaryCategories::WasmReactor => false,
```
sinkingsugar: for now do `infer::is(data, "wasm")`

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+
+	/// Is the call `c` valid?
+	///
+	/// Note: This function does not check whether the child/descendant calls of `c` (if it has any) are valid.
+	pub fn is_the_immediate_call_valid(c: &Call) -> bool {
+		match c {
+			Call::Protos(ProtosCall::upload{ref data, ref category, ref references, ..}) => {
+				match data {
+					pallet_protos::ProtoData::Local(ref data) => is_valid(category, data, references),
+					_ => false, // TODO Review
```
sinkingsugar: true

runtime/src/lib.rs
```diff
@@ -238,11 +358,453 @@ parameter_types! {
...
+				}
+			},
+			Call::Protos(ProtosCall::patch{ref proto_hash, ref data, ref new_references, ..}) => {
+				let Some(proto_struct) = pallet_protos::Protos::<Runtime>::get(proto_hash) else {
+					return false;
+				};
+				match data {
+					None => true,
+					Some(pallet_protos::ProtoData::Local(ref data)) => is_valid(&proto_struct.category, data, new_references),
+					_ => false, // TODO Review
```
sinkingsugar: true

sinkingsugar (changes requested):

sinkingsugar:
@aang114 
I have a request before merging also:
We need to enforce a few categories to be `Local` only.
Those are:
- [ ] Shards
- [ ] Traits
- [ ] Text

aang114:
> @aang114 I have a request before merging also: We need to enforce a few categories to be `Local` only. Those are:
> 
> * [ ]  Shards
> * [ ]  Traits
> * [ ]  Text

I'm not sure what you mean? What does `Local` mean?

runtime/src/lib.rs
aang114: That file extension doesn't exist in the `infer` crate

aang114: > That file extension doesn't exist in the `infer` crate

so I'll just put `true` for now

aang114:
> @aang114 I have a request before merging also: We need to enforce a few categories to be `Local` only. Those are:
> 
> * [ ]  Shards
> * [ ]  Traits
> * [ ]  Text

done @sinkingsugar

runtime/src/lib.rs
```diff
@@ -238,11 +358,460 @@ parameter_types! {
...
+					})
+				});
+
+				let all_traits_implemented_in_this_shards = implementing.iter().all(|_shards_trait| {
+					match format {
+						ShardsFormat::Edn => {
+							true
+						},
+						ShardsFormat::Binary => {
+							true
```
sinkingsugar: let's do false for now

runtime/src/lib.rs
```diff
@@ -238,11 +358,460 @@ parameter_types! {
...
+			Categories::Vector(sub_categories) => match sub_categories {
+				VectorCategories::SvgFile => false,
+				VectorCategories::TtfFile => infer::is(data, "ttf"), // ttf_parser::Face::parse(&data[..], 0).is_ok(),
+			},
+			Categories::Video(sub_categories) => match sub_categories {
+				VideoCategories::MkvFile => infer::is(data, "mkv"),
+				VideoCategories::Mp4File => infer::is(data, "mp4"),
+			},
+			Categories::Model(sub_categories) => match sub_categories {
+				ModelCategories::GltfFile => true,
```
sinkingsugar: this makes this whole PR worthless, let's do false 👍

sinkingsugar (changes requested):
just some final things

runtime/src/lib.rs
aang114: for everything I just made `true`, right?

sinkingsugar: what do you mean?
I mean a malicious party can select gltf and upload anything.

aang114: okay nevermind. I thought you wanted me to revert all the comments earlier where you said "put true". So I'll just make this `false`

aang114: done

sinkingsugar (approved):
CI green and do a big formatting pass cargo fmt or so before merge please

PR #209, opened 2022-10-14
Allow Secondary Sale of Fragment Instances

aang114:


pallets/fragments/src/lib.rs
```diff
@@ -304,6 +327,10 @@ pub mod pallet {
 	pub type Publishing<T: Config> =
 		StorageMap<_, Identity, Hash128, PublishingData<T::BlockNumber>>;
 
+	#[pallet::storage]
+	pub type Definition2SecondarySales<T: Config> =
```
aang114: @sinkingsugar , **instead of it being** 
```
#[pallet::storage]
pub type Definition2SecondarySales<T: Config> =
StorageMap<_, Identity, Hash128, Vec<SecondarySaleData<T::AccountId, T::BlockNumber>>>;
```

**why can't we have it as:**

```
#[pallet::storage]
pub type Definition2SecondarySales<T: Config> = StorageNMap<
_,
(
	// Definition Hash
	storage::Key<Identity, Hash128>,
	// Edition ID
	storage::Key<Identity, Unit>,
	// Copy ID
	storage::Key<Identity, Unit>,
),
SecondarySaleData<T::AccountId, T::BlockNumber>,
>;
```

The storage used for to store an instance's `SecondarySaleData` would be the same. But by using `StorageNMap` it gives us the added benefit of querying and deleting an instance's `SecondarySaleData` in O(1) time.

Therefore, please let me know if we should use a `StorageNMap` instead. Thanks

sinkingsugar: sounds good to me

aang114: Okay. Thanks

pallets/fragments/src/tests.rs
```diff
@@ -2100,6 +2100,11 @@ mod create_account_tests {
 	}
 }
 
+mod resell_tests {
```
aang114: @alebaffa @sinkingsugar, what is our "testing philosophy" 😅 . Must we only want to test "public functions" and never "private functions"?

Because if you notice that `fragments.resell()` uses the same private function `fragments.can_transfer_instance()` as `fragments.give()` - so a lot of the test cases will be similar. for example: `give_tests.give_should_work_if_the_expiration_parameter_is_greater_than_the_current_block_number()`, `give_test.give_should_not_work_if_the_new_permissions_are_less_restrictive()` etc. etc...

So I don't wanna waste time writing the exact same tests in `resell_tests`. But I'm not sure what's the correct convention/approach 😅?

alebaffa: Everything needs to be tested. Of course no need to replicate tests. If a private function is already tested once, that's enough.

aang114:
This PR does the following things:
1. Enables secondary sale of instances by adding the extrinsics: `fragments.resell()`, `fragments.unresell()` and `fragments.secondary_buy()`, along with their respective unit test cases (THIS IS THE MAIN THING THIS PR DOES)
2. In the extrinsic `fragments.create()`, **previously the created definition vault account was not created if `currency` was not None**. This has now been fixed in this PR
3. Refactored some existing unit test cases 

(btw @sinkingsugar , @alebaffa  - do you have a better name to suggest than `resell()`, `unresell()` (😂 ) and `secondary_buy()`. Thanks)

pallets/fragments/src/lib.rs
```diff
@@ -1066,256 +1106,227 @@ pub mod pallet {
...
+		/// End the sale of a Fragment Instance that is currently on sale.
+		///
+		/// # Arguments
+		///
+		/// * `origin` - **Origin** of the **extrinsic function**
+		/// * `definition_hash` - Fragment Definition ID of the Fragment Instance
+		/// * `edition` - Edition ID of the Fragment Instance
+		/// * `copy` - Copy ID of the Fragment instance
+		#[pallet::weight(50_000)]
+		pub fn unresell(
```
alebaffa: `unresell` is really not intuitive at all as a name 😄 
Let's use a meaningful and self-explanatory name that can be as simple as `end_sale`, for example. 
What do you think?

pallets/fragments/src/lib.rs
```diff
@@ -1538,6 +1549,277 @@ pub mod pallet {
...
+
+			if let Some(new_perms) = new_permissions {
+				// ensure we only allow more restrictive permissions
+				if (item_data.permissions & FragmentPerms::EDIT) != FragmentPerms::EDIT {
+					ensure!(
+						(new_perms & FragmentPerms::EDIT) != FragmentPerms::EDIT,
+						Error::<T>::NoPermission
+					);
+				}
+				if (item_data.permissions & FragmentPerms::COPY) != FragmentPerms::COPY {
```
alebaffa, line 1692: Shouldn't be `else if` in the 2nd and 3rd statements?

alebaffa (approved):

pallets/fragments/src/lib.rs
sinkingsugar: `stop_resellling` explicit?

pallets/fragments/src/lib.rs
```diff
@@ -1066,256 +1106,227 @@ pub mod pallet {
...
+		/// * `new_permissions` (*optional*) - The permitted set of actions that the account that is given the Fragment instance can do with it. Note: `new_permissions` must be a subset of the current `permissions` field of the Fragment Instance.
+		/// * `expiration` (*optional*) - Block number that the newly-copied Fragment Instance expires at. If the Fragment Instance is not copyable, this parameter is practically irrelevant.
+		/// * `secondary_sale_type` - Type of Sale
+		#[pallet::weight(50_000)]
+		pub fn resell(
+			origin: OriginFor<T>,
+			definition_hash: Hash128,
+			edition_id: Unit,
+			copy_id: Unit,
+			new_permissions: Option<FragmentPerms>,
```
sinkingsugar, line 1234: are we checking that new perms are compatible with existing ones and more restrictive or equal?

sinkingsugar (dismissed):

pallets/fragments/src/lib.rs
aang114: what about `stop_resale`?

pallets/fragments/src/lib.rs
aang114: of course. that's we we call `Self::can_transfer_instance` in this extrinsic

pallets/fragments/src/lib.rs
aang114: the 3 if statements are:
- if `item.permissions` doesn't contain `FragmentPerms::EDIT`, then `new_perms` shouldn't either
- if `item.permissions` doesn't contain `FragmentPerms::COPY`, then `new_perms` shouldn't either
- if `item.permissions` doesn't contain `FragmentPerms::TRANSFER`, then `new_perms` shouldn't either

pallets/fragments/src/lib.rs
aang114: nvm reselling is better

aang114:
@sinkingsugar @alebaffa , can I merge this? Thanks

alebaffa:
LGTM 👍🏻

PR #211, opened 2022-10-20
Fix the CI's IPFS test + Update the Substrate Tag to `fragnova-v0.9.30` in Clamor 

aang114:


shards/run-test-ipfs.edn
```diff
@@ -0,0 +1,89 @@
...
+  "" (Process.Run "ipfs" :Arguments .ipfs-args :Timeout 120) (StringToBytes) ; `ipfs block` - Interact with raw IPFS blocks. A block is identified by a Multihash passed with a valid CID.
+
+)
+
+(defwire main
+  (Setup
+   (get Fragnova-Node :setup)
+  )
+
+  "Proto-Indo-European" (ToBytes) = .proto-data
```
aang114, line 74: @sinkingsugar @Kryptos-FR @guusw @tdbe @alebaffa @ShahFragcolor @ChelseaChX  - Giovanni, like I told you earlier, since the **current wire's variables are exposed to a child wire** whenever a child wire is called (i.e whenever `(Do <child wire name>)` is called) - **it makes it very difficult and cumbersome to use Shards "Input Output paradigm"** because a **shards scripter will always be worrying whether the variable name he initialized in the child wire already exists in the wire that calls the child wire!!!!!!!!**

For example, if you look at this example:

**PARENT WIRE:**
```clojure
"Proto-Indo-European" (ToBytes) (Do upload-proto)
**CHILD WIRE:**
```clojure
(defwire upload-proto
  (ExpectBytes) = **.proto-data**
  UPLOAD-CALL-DATA-EXCL-DATA-PARAM (HexToBytes) >= .call-data
  .proto-data >> .seq .seq (Substrate.Encode [nil]) (AppendTo .call-data)
  .call-data (Do (get Fragnova-Node :send-signed-extrinsic-as-http-request)) (Log "HTTP Response of protos.upload() is")
  .proto-data (Hash.Blake2-256) (| (ToHex) (Log "Uploaded Proto With Proto Hash"))
)
```

If the variable `.proto-data` in the CHILD WIRE (i.e in `upload-proto`) was initialized in the PARENT WIRE - then Shards will not even work - which is very frustrating!!!

**That's why I'm forced to write it like this in my code (which is extremely cumbersome) to "be safe":**
**PARENT WIRE:**
```clojure
  "Proto-Indo-European" (ToBytes) = .proto-data
  (Do upload-proto) ; `upload-proto` expects a variable called `proto-data`
```
**CHILD WIRE:**
```clojure
(defwire upload-proto
  **.proto-data** (ExpectBytes)
  UPLOAD-CALL-DATA-EXCL-DATA-PARAM (HexToBytes) >= .call-data
  .proto-data >> .seq .seq (Substrate.Encode [nil]) (AppendTo .call-data)
  .call-data (Do (get Fragnova-Node :send-signed-extrinsic-as-http-request)) (Log "HTTP Response of protos.upload() is")
  .proto-data (Hash.Blake2-256) (| (ToHex) (Log "Uploaded Proto With Proto Hash"))
)
```
But this approach is extremely cumbersome and doesn't utilize Shard's Input-Ouput Paradigm!

alebaffa: Is this the reason why the IPFS tests fail with the new Substrate tag?

alebaffa:
Thanks @aang114 for this! I have integrated your updates made here into another PR https://github.com/fragcolor-xyz/clamor/pull/213 to keep things clean in terms of git history.

So, I am gonna close this PR.

PR #215, opened 2022-11-02
Added extrinsic `fragments.detach()` and added tests for `pallet-detach`

aang114:


aang114:
The following things were done @sinkingsugar @alebaffa :
- Used an enum named `DetachHash` in `pallet-detach` to represent the different kind of "things that can be detached" (for now it's only Proto-Fragment and Fragment Instance)
- Added a separate crypto key type for `pallet-detach` (not sharing `b"frag"` anymore) 
- Added tests for `pallet-detach`
- Refactored `pallet-detach`
   - `StorageValueRef::persistent(b"fragments-detach-requests")` was renamed to `StorageValueRef::persistent(b"detach-requests")`
   - `StorageValueRef::persistent(b"fragments-frag-ecdsa-keys")` was renamed to `StorageValueRef::persistent(b"detach-ecdsa-keys")`
   - made the pallet easier to read by adding helper functions
- Added documentation for `pallet-detach`
- Added extrinsic `fragments.detach()` for detaching Fragment Instances

pallets/protos/src/lib.rs
```diff
@@ -796,10 +798,14 @@ pub mod pallet {
 				},
 			};
 
-			ensure!(!<DetachedHashes<T>>::contains_key(&proto_hash), Error::<T>::Detached);
+			let detach_hash = DetachHash::Proto(proto_hash);
+			let detach_request = DetachRequest { hash: detach_hash.clone(), target_chain, target_account };
+
+			ensure!(!<DetachedHashes<T>>::contains_key(&detach_hash), Error::<T>::Detached);
+			ensure!(!<DetachRequests<T>>::get().contains(&detach_request), Error::<T>::DetachRequestAlreadyExists);
```
aang114, line 805: @sinkingsugar is there any need to check to check if the detach requests already exists. I don't think there is, but I just added this line to be overly cautious...because why not 😅

pallets/fragments/src/lib.rs
```diff
@@ -1330,6 +1334,61 @@ pub mod pallet {
...
+				.ok_or(Error::<T>::NotFound)?;
+			if let Some(expiring_at) = instance.expiring_at {
+				ensure!(current_block_number < expiring_at, Error::<T>::NotFound);
+			}
+
+			let detach_hash = DetachHash::Instance(definition_hash, Compact(edition_id), Compact(copy_id));
+			let detach_request = DetachRequest { hash: detach_hash.clone(), target_chain, target_account};
+
+			ensure!(!<DetachedHashes<T>>::contains_key(&detach_hash),Error::<T>::Detached);
+			ensure!(!<DetachRequests<T>>::get().contains(&detach_request), Error::<T>::DetachRequestAlreadyExists);
```
aang114, line 1394: @sinkingsugar is there any need to check to check if the detach requests already exists. I don't think there is, but I just added this line to be overly cautious...because why not 😅

pallets/detach/src/lib.rs
```diff
@@ -443,243 +510,256 @@ pub mod pallet {
...
-		/// NOTE: `Signer::<T, T::AuthorityId>::any_account()` uses any of the keys that was added using the RPC `author_insertKey` into Clamor (https://polkadot.js.org/docs/substrate/rpc/#insertkeykeytype-text-suri-text-publickey-bytes-bytes)
-		fn process_detach_requests() {
+		/// 2. The Detach-Nonce of the detach request's target account. (See `DetachNonces` for more information).
+		fn get_detach_signature_and_detach_nonce(request: &DetachRequest) -> Result<(Vec<u8>, u64), Error<T>> {
+			match request.target_chain {
+				SupportedChains::EthereumMainnet
+				| SupportedChains::EthereumRinkeby
+				| SupportedChains::EthereumGoerli => {
+
+					Self::add_newly_found_ed25519_and_ecdsa_keys();
```
aang114, line 591: @sinkingsugar why should we check for new Ed25519 keys for each iteration of `StorageValueRef::persistent(b"detach-requests")`. shouldn't we only check it once at the beginning of `detach.process_detach_requests()`?

Here's how it is done in the current code for reference (my question might make more sense when you see this 😅):
https://github.com/fragcolor-xyz/clamor/blob/devel/pallets/detach/src/lib.rs#L485-L514

pallets/detach/src/lib.rs
```diff
@@ -443,243 +510,256 @@ pub mod pallet {
...
+					let mut chain_id_be: [u8; 32] = [0u8; 32]; // "be" stands for big-endian
+					match request.target_chain {
+						SupportedChains::EthereumMainnet => U256::from(1),
+						SupportedChains::EthereumRinkeby => U256::from(4),
+						SupportedChains::EthereumGoerli => U256::from(5),
+					}.to_big_endian(&mut chain_id_be);
+
+					let payload = [
+						// REVIEW - `request.hash` is a `DetachHash` enum. Do we want that? @sinkingsugar
+						&request.hash.encode()[..],
```
aang114: The payload's (which will then signed) first piece of data is `request.hash.encode()`, where `request.hash` is a `DetachHash` enum. 

Do we want that? @sinkingsugar 

Because if we do it like this, we would have to **pass the same SCALE encoded  `DetachHash` enum as a parameter to our EVM Smart Contract's "attach" function** when attaching the Proto-Fragment / Fragment Instance.

aang114: Actually it would be annoying since we would have to decipher the SCALE encoded Rust enum to either a Proto-Fragment or a Fragment Instance in our EVM Smart Contract to understand if it's a Proto or a FI!

I think it would be better to just add the Proto-Fragment's Hash or the Fragment Instance's Tuple into the payload then right? and we can just concat the string "proto" to the payload if we wanna detatch/attach a Proto-Fragment or concat the string "instance" to the payload if we wanna detach/attach a Fragment Instance?

alebaffa: Personally, I would avoid to have to decipher the SCALE encoded Rust enum. At the moment our smart contract receives the hash, I would keep it simple.

alebaffa (approved):

pallets/detach/src/lib.rs
aang114: > Actually it would be annoying since we would have to decipher the SCALE encoded Rust enum to either a Proto-Fragment or a Fragment Instance in our EVM Smart Contract to understand if it's a Proto or a FI!
> 
> I think it would be better to just add the Proto-Fragment's Hash or the Fragment Instance's Tuple into the payload then right? and we can just concat the string "proto" to the payload if we wanna detatch/attach a Proto-Fragment or concat the string "instance" to the payload if we wanna detach/attach a Fragment Instance?

so this is a good solution, right? @alebaffa @sinkingsugar ?

aang114: Okay I've simplified it (just pushed) @alebaffa

pallets/detach/src/lib.rs
aang114: Here's the updated link: https://github.com/fragcolor-xyz/clamor/pull/215/files#diff-f3de07c470ea31d8f5f62b2949c157b04d3506ecbf23fe6daf37898f05769499L485-L514

PR #224, opened 2022-11-14
Add pallet Clusters

alebaffa:
This PR introduces the pallet-clusters.


node/src/command.rs
```diff
@@ -8,7 +8,8 @@
 use crate::{
```
alebaffa: This file should not be here in this PR. I will delete it. Mistakenly taken because of a local `cargo fmt`.

alebaffa:
One question @sinkingsugar 
- `a cluster can have an account or multiple accounts` does it mean that a Cluster can have multiple `AccountId `(so multiple balances) associated to it?

pallets/clusters/src/lib.rs
```diff
@@ -29,40 +29,50 @@ pub struct ClusterMember {
...
 	pub data: Vec<u8>,
 }
 
 /// **Struct** of **Role** belonging to a **Cluster**.
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
 pub struct Role {
+	/// Name of the role
 	pub name: Vec<u8>,
+	/// Settings of the Role
 	pub settings: RoleSettings,
```
sinkingsugar: can't we just use a map?

pallets/clusters/src/lib.rs
```diff
@@ -29,40 +29,50 @@ pub struct ClusterMember {
...
-	pub members: Vec<ClusterMember>,
+	/// The list of Members associated to the role
```
sinkingsugar:
```suggestion
	/// The list of Members associated with the role
```

pallets/clusters/src/lib.rs
```diff
@@ -142,16 +152,17 @@ pub mod pallet {
 		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
 		pub fn create_cluster(origin: OriginFor<T>, name: Vec<u8>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
-			ensure!(!name.len().is_zero(), Error::<T>::SystematicFailure);
+			ensure!(!name.len().is_zero(), Error::<T>::InvalidInputs);
 
 			let cluster_hash = blake2_256(&name);
-
+			// Check that the cluster does not exist already
 			ensure!(!<Clusters<T>>::contains_key(&cluster_hash), Error::<T>::ClusterExists);
```
sinkingsugar: I'm actually unsure about this, we might want clusters with the same name.
I think the cluster can be a 128 bit hash, and a random hash is fine (pseudo, mix e.g. (extrinsic index+origin+name))

sinkingsugar: We want to add Cluster ID to aliases types too

pallets/clusters/src/lib.rs
```diff
@@ -142,16 +152,17 @@ pub mod pallet {
...
 
 			// At creation there are no roles and no members assigned to the cluster
 			let cluster = Cluster { owner: who.clone(), name, roles: vec![], members: vec![] };
 
 			// create an account for the cluster
+			// TODO give the possibility to cluster's owner to add multiple accounts to the cluster
```
sinkingsugar: I mean via proxy etc they can already

pallets/clusters/src/lib.rs
```diff
@@ -29,40 +29,50 @@ pub struct ClusterMember {
...
 
 /// **Struct** of a **Cluster**
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
 pub struct Cluster<TAccountId> {
+	/// The owner of the Cluster
 	pub owner: TAccountId,
+	/// The name of the Cluster
 	pub name: Vec<u8>,
+	/// The list of Roles associated to the Cluster
 	pub roles: Vec<Hash256>,
```
sinkingsugar: Hmm I'd say roles need to be embedded, not in a different structure.

pallets/clusters/src/lib.rs
```diff
@@ -29,40 +29,50 @@ pub struct ClusterMember {
...
+	/// The list of Members of the Cluster
 	pub members: Vec<Hash256>,
```
sinkingsugar: what's hash256?, this should be AccountID no?

pallets/clusters/src/lib.rs
```diff
@@ -221,26 +237,58 @@ pub mod pallet {
...
 			ensure!(roles_in_cluster.contains(&role_hash), Error::<T>::RoleNotFound);
 			ensure!(<Roles<T>>::contains_key(role_hash), Error::<T>::RoleNotFound);
 
+			let current_settings =
+				<Roles<T>>::get(role_hash).ok_or(Error::<T>::SystematicFailure)?.settings;
+
+			// If there are no new settings, override. Otherwise use the old ones.
+			let settings = if let Some(new_settings) = new_settings {
+				if !new_settings.data.is_empty() && !new_settings.name.is_empty() {
+					new_settings
```
sinkingsugar: can you batch remove settings?

pallets/clusters/src/lib.rs
```diff
@@ -318,13 +375,57 @@ pub mod pallet {
...
 			Ok(())
 		}
 
+		/// Edit the data of a **Member** in an existing cluster.
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		pub fn edit_member(
+			origin: OriginFor<T>,
+			cluster_hash: Hash256,
+			member_hash: Hash256,
+			new_member_data: Option<Vec<u8>>,
```
sinkingsugar: we are allowing arbitrary lengths of data to be stored inside the actual chain state here.. this is not good.
I just realised we have many of those `Vec<u8>` they all need to be replaced with BoundedVec, whenever the data is STORED in the live state. (See protos upload, data IS NOT STORED anywhere that's why unbound is ok)
Let's make a story about this asap as we are increasing block size and capabilities soon, and this refactor needs to come first.

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
alebaffa: A random hash brings in the problem of how to retrieve it from other functions later on. For example, when the owner of the Cluster wants to create a new Role and add it to an existing Cluster (which is empty at the beginning, no Members and no Roles), he needs a way to retrieve the hash of that Cluster. But if there are multiple clusters with the same name and same owner, there is no way to recreate the hash of the exact Cluster he wants to add the Role to.
Also Protos and Fragments have a hash that can be generated from the data they contain, it's not random.

So, I am thinking to change the returned value of the extrinsic create_cluster to `DispatchResultWithPostInfo` so that the random cluster_id generated can be returned to caller and be used with the other extrinsics (e.g.` create_role`, `add_member`, etc...). It would be a first time in our codebase, though.

pallets/clusters/src/lib.rs
alebaffa: I removed `RoleSetting` from the `Role` structure and added a storage [`RoleToSettings`](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L112).

pallets/clusters/src/lib.rs
alebaffa: Good idea! Updated it to be [random](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L177) and also created a [cluster_id](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L58) field that is used by the other extrinsic and will be used for the pallet-aliases too.

pallets/clusters/src/lib.rs
alebaffa: Added the possibility for the cluster's owner to [add proxy accounts](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L429) to the account of the cluster.

pallets/clusters/src/lib.rs
alebaffa: Right. Updated the cluster struct to embed roles as [`pub roles: Vec<Role<TAccountId>>`](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L60).

pallets/clusters/src/lib.rs
alebaffa: It was `Vec<Hash256>` because I had a dedicated struct for `Member` that also included `data`, but now I removed it because it is not clear what a `data` field could mean for a member at the moment (also considering that it was a `Vec<u8>`, so it also needed a limit).
So, now the cluster has the list of members as [`Vec<T::AccountId>`](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L62).

pallets/clusters/src/lib.rs
alebaffa: Ok. So now if the cluster's owner sends an empty list of role settings, it will [delete the existing ones](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L302).
Same in case the list of [members passed as input is empty](https://github.com/fragcolor-xyz/clamor/blob/8276f14d8e8d6eb42a1fcc163adbc348619888ed/pallets/clusters/src/lib.rs#L291).

pallets/clusters/src/lib.rs
alebaffa: Agreed. As said before, I removed the Member struct, so there is no more `member_data: Vec<u8>` field since it is not clear what it means and consequently difficult to find a limit for it at the moment.

node/src/command.rs
sinkingsugar: let's not forger this

pallets/clusters/README.md
```diff
@@ -0,0 +1 @@
+License: Unlicense
```
sinkingsugar: should be BUSL or something

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+mod dummy_data;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+
+/// The **settings** of a **Role**
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
+pub struct RoleSetting {
+	/// Name of the setting
+	pub name: Vec<u8>,
```
sinkingsugar, line 29: we decided to refactor bounded vectors after?
Maybe we should start here though (in the scope of clusters)?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
+				.ok_or(Error::<T>::SystematicFailure)?;
+
+			let cluster_id = blake2_128(
+				&[name.clone(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
+			);
+
+			sp_std::if_std! {
+				// This code is only being compiled and executed when the `std` feature is enabled.
+				println!("CLUSTER_ID: {:?}", cluster_id);
```
sinkingsugar: println is wrong, you should use log crate as it can be useful and easily enabled/disabled via env

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		pub fn create_cluster(origin: OriginFor<T>, name: Vec<u8>) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(!name.len().is_zero(), Error::<T>::InvalidInputs);
+
+			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
+				.ok_or(Error::<T>::SystematicFailure)?;
+
+			let cluster_id = blake2_128(
+				&[name.clone(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
```
sinkingsugar: Maybe add block number, otherwise same account cannot create same name cluster

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+
+			Ok(())
+		}
+
+		/// Allow another Account ID `proxy_account` to be used as a proxy for the Account ID of the cluster
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1).ref_time())]
+		pub fn add_proxy_account(
+			origin: OriginFor<T>,
+			cluster_id: Hash128,
+			proxy_account: T::AccountId,
```
sinkingsugar: I don't get it, I'd rather just create a proxy here, not add the delegate directly (not the proxy in fact)

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+			// Check that the caller is the owner of the cluster
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			ensure!(
+				!pallet_proxy::Proxies::<T>::contains_key(&proxy_account),
+				Error::<T>::AccountProxyAlreadyExists
+			);
+
+			let cluster_account = Self::get_vault_id(cluster_id);
```
sinkingsugar: maybe get_vault_it should be a common utility in sp-clamor

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
alebaffa: oh damn, sorry! Forgot to remove it! just a quick super dirty way to see things in log when I make tests.

pallets/clusters/src/lib.rs
```diff
@@ -22,42 +26,42 @@ mod benchmarking;
 
 /// The **settings** of a **Role**
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
-pub struct RoleSetting {
+pub struct RoleSetting<BoundedName, BoundedData> {
 	/// Name of the setting
-	pub name: Vec<u8>,
+	pub name: BoundedName,
```
sinkingsugar: I think all those should rather be `BoundedVec<u8, TNameLimit>` vec u8 doesn't ever change.

alebaffa: I tried that and I get a lots of errors due to missing codecs and implementations of macro. I spent two days on it.
I will try again and post here the errors I get.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+	pub data: BoundedData,
+}
+
+/// **Struct** of **Role** belonging to a **Cluster**.
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
+pub struct Role<TAccountId, BoundedName> {
+	/// Name of the role
+	pub name: BoundedName,
+	/// The list of Members associated with the role
+	pub members: Vec<TAccountId>,
```
sinkingsugar, line 40: We want to set a hard limit here too

pallets/clusters/src/lib.rs
sinkingsugar: It's the Get you are using for the config that misses `WrapperTypeEncode`, which is fine to use, leave it like this, it's minor.
You could have simply explained this btw also no need to spend two days on such issues 🙏

node/src/command.rs
alebaffa: removed.

pallets/clusters/README.md
alebaffa: Fixed.

pallets/clusters/src/lib.rs
alebaffa: As discussed, `BoundedVec` now are used only in the extrinsics input, not in storage nor in structs.

pallets/clusters/src/lib.rs
alebaffa: Added.

pallets/clusters/src/lib.rs
alebaffa: Proxy now is created inside the function [create_cluster](https://github.com/fragcolor-xyz/clamor/pull/224/files#diff-6804b4445503eaa0a518d6cabdb56e2901be2c3b0c19ba2b7b4a5e4becdad173R233) like so:
```rust
// create a pure proxy
pallet_proxy::Pallet::<T>::create_pure(
	origin.clone(),
	T::ProxyType::default(),
	T::BlockNumber::zero(),
	0,
)?;
```

pallets/clusters/src/lib.rs
alebaffa: Done. Moved in [sp-clamor](https://github.com/fragcolor-xyz/clamor/blob/749c5c4f68abccbeb7de0e580c90c00d747f4c84/primitives/clamor/src/lib.rs#L221).

pallets/clusters/src/lib.rs
```diff
@@ -165,38 +169,40 @@ pub mod pallet {
...
-			ensure!(!bounded_name.len().is_zero(), Error::<T>::InvalidInputs);
+			ensure!(!name.len().is_zero(), Error::<T>::InvalidInput);
 
 			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
 				.ok_or(Error::<T>::SystematicFailure)?;
 			let current_block_number = <frame_system::Pallet<T>>::block_number();
 
 			let cluster_id = blake2_128(
-				&[current_block_number.encode(), name.clone(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
+				&[current_block_number.encode(), name.to_vec().clone(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
```
sinkingsugar: eh, this is more allocations etc... can't we access the underlying one? `&name` `name.0` something?

pallets/clusters/src/lib.rs
```diff
@@ -205,17 +211,20 @@ pub mod pallet {
 			// At creation there are no roles and no members assigned to the cluster
 			let cluster = Cluster {
 				owner: who.clone(),
-				name: bounded_name,
+				name: name.to_vec(),
```
sinkingsugar: more copies...

pallets/clusters/src/lib.rs
```diff
@@ -243,34 +252,32 @@ pub mod pallet {
...
-			ensure!(!bounded_name.len().is_zero(), Error::<T>::InvalidInputs);
-			ensure!(!settings.name.len().is_zero(), Error::<T>::InvalidInputs);
-			ensure!(!settings.data.len().is_zero(), Error::<T>::InvalidInputs);
+			ensure!(!cluster_id.len().is_zero(), Error::<T>::InvalidInput);
+			ensure!(!role_name.len().is_zero(), Error::<T>::InvalidInput);
+			ensure!(!settings.name.len().is_zero(), Error::<T>::InvalidInput);
+			ensure!(!settings.data.len().is_zero(), Error::<T>::InvalidInput);
 
-			let role_hash = blake2_128(&[&cluster_id[..], &bounded_name.clone()[..]].concat());
+			let role_hash = blake2_128(&[&cluster_id[..], &role_name.to_vec().clone()[..]].concat());
```
sinkingsugar: double copy in 1 line ? to_vec + clone 👎

pallets/clusters/src/lib.rs
```diff
@@ -243,34 +252,32 @@ pub mod pallet {
...
 
 			// Check that the caller is the owner of the cluster
 			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
 			ensure!(who == cluster.owner, Error::<T>::NoPermission);
 
 			// At creation there are no Members and no Rules assigned to a Role
-			let role = Role { name: bounded_name.clone(), members: vec![], rules: None };
+			let role = Role { name: role_name.to_vec().clone(), members: vec![], rules: None };
```
sinkingsugar: copy

pallets/clusters/src/lib.rs
```diff
@@ -243,34 +252,32 @@ pub mod pallet {
...
 
 			// Check that the role does not exists already in the cluster
 			let roles_in_cluster = <Clusters<T>>::get(&cluster_id)
 				.ok_or(Error::<T>::ClusterNotFound)?
 				.roles
 				.into_iter()
-				.filter(|role| role.name == bounded_name)
-				.collect::<Vec<Role<T::AccountId, BoundedVec<u8, T::NameLimit>>>>();
+				.filter(|role| role.name == role_name.to_vec())
```
sinkingsugar: more copies just to compare the value... do by ref

pallets/clusters/src/lib.rs
```diff
@@ -316,16 +322,16 @@ pub mod pallet {
...
```
sinkingsugar: copy to compare 👎

pallets/clusters/src/lib.rs
```diff
@@ -316,16 +322,16 @@ pub mod pallet {
...
+				.collect::<Vec<Role<T::AccountId>>>();
 			ensure!(!roles_in_cluster.is_empty(), Error::<T>::RoleNotFound);
 
-			let role_hash = blake2_128(&[&cluster_id[..], &bounded_name.clone()].concat());
+			let role_hash = blake2_128(&[&cluster_id[..], &role_name.to_vec().clone()].concat());
```
sinkingsugar: double copy

pallets/clusters/src/lib.rs
```diff
@@ -316,16 +322,16 @@ pub mod pallet {
...
 
 			// write
 			<Clusters<T>>::mutate(&cluster_id, |cluster| {
 				let cluster = cluster.as_mut().unwrap();
-				let index = cluster.roles.iter().position(|x| x.name == bounded_name);
+				let index = cluster.roles.iter().position(|x| x.name == role_name.to_vec());
```
sinkingsugar: copy to compare

sinkingsugar (changes requested):
I could have gone on as there are many repetitions of copies.

pallets/clusters/src/lib.rs
alebaffa: Oh, I did not realize `to_vec` was creating a copy. Thank you!
I changed it everywhere now to use `Vec::from(bounded_vec)`, which underneath does 
```rust
fn from(x: BoundedVec<T, S>) -> Vec<T> {
	x.0
}
```
 I could not do `bounded_vec.0` myself since it threw the error `0 is a private field`.

I used it as a reference everywhere I could. Much cleaner now.
Many thanks!

pallets/clusters/src/lib.rs
```diff
@@ -202,7 +202,7 @@ pub mod pallet {
 			let current_block_number = <frame_system::Pallet<T>>::block_number();
 
 			let cluster_id = blake2_128(
-				&[current_block_number.encode(), name.to_vec().clone(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
+				&[current_block_number.encode(), Vec::from(name.clone()), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
```
sinkingsugar: This changes nothing.
please use https://doc.rust-lang.org/std/slice/trait.Concat.html on slices, use `name.as_slice()` and encode is not needed for things that are already bytes etc.
@aang114 can you have a look please, we need to be max efficient here

pallets/clusters/src/lib.rs
```diff
@@ -262,21 +262,22 @@ pub mod pallet {
 			ensure!(!settings.name.len().is_zero(), Error::<T>::InvalidInput);
 			ensure!(!settings.data.len().is_zero(), Error::<T>::InvalidInput);
 
-			let role_hash = blake2_128(&[&cluster_id[..], &role_name.to_vec().clone()[..]].concat());
+			let role_name_vec = Vec::from(role_name.clone());
```
sinkingsugar: `as_slice`

pallets/clusters/src/lib.rs
alebaffa: I already tried to use .as_slice, but I did not use it because it causes

```rust
 &[current_block_number.encode(), name.as_slice(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
      |                                                  ^^^^^^^^^^^^^^^- help: try using a conversion method: `.to_vec()`
      |                                                  |
      |                                                  expected struct `frame_support::dispatch::Vec`, found `&[u8]`

```
I can use `Vec::from(name.clone().as_slice())` instead.

pallets/clusters/src/lib.rs
alebaffa: I already tried to use .as_slice(), but I did not use because it cause:
```rust
--> /home/workspace/fragcolor/clamor/pallets/clusters/src/lib.rs:205:38
      |
  205 |                 &[current_block_number.encode(), name.as_slice(), extrinsic_index.clone().encode(), who.clone().encode()].concat(),
      |                                                  ^^^^^^^^^^^^^^^- help: try using a conversion method: `.to_vec()`
      |                                                  |
      |                                                  expected struct `frame_support::dispatch::Vec`, found `&[u8]`
      |
      = note: expected struct `frame_support::dispatch::Vec<u8>`
              found reference `&[u8]`
```

pallets/clusters/src/lib.rs
sinkingsugar: Ah I see that this passes the inner directly ok, can you comment that, I miseed from code review

pallets/clusters/src/lib.rs
sinkingsugar: Ok changes something, missed the Vec::from part, but still there are a lot of allocations around

alebaffa: Those `encode` are all needed because none of the variables passed in blake2_128 (`BlockNumber`, `u32`, `AccountId`) are bytes, as far as I can see.
I removed one `.clone()` from a `log.trace` that was not needed, but all the other allocations (i.e. `clone()`) are used where a reference could not be passed.
I don't see how can I optimize more. Maybe @aang114 can spot something more?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+			let who = ensure_signed(origin.clone())?;
+
+			ensure!(!name.len().is_zero(), Error::<T>::InvalidInput);
+
+			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
+				.ok_or(Error::<T>::SystematicFailure)?;
+			let current_block_number = <frame_system::Pallet<T>>::block_number();
+
+			let cluster_id = blake2_128(
+				&[current_block_number.encode(), Vec::from(name.clone().as_slice()), extrinsic_index.clone().encode(), who.encode()].concat(),
```
sinkingsugar: `Vec::from(name.clone().as_slice())` what is this abomination?

sinkingsugar:
@alebaffa please apply knowledge from this https://github.com/fragcolor-xyz/clamor/pull/224/commits/133b4df030efb282d593dfedfa25abe6072e1357 to the rest.

pallets/clusters/src/lib.rs
alebaffa: I removed `Vec::from(bounded_vec)` and used `bounded_vec.into_inner()`, which even simpler does `self.0` and it is provided by bounded_vec itself without the need to pass through Vec.

I applied other fixes too.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+
+			// Check that the caller is the owner of the cluster
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			// At creation there are no Members and no Rules assigned to a Role
+			let new_role = Role { name: role_name.into_inner(), members: vec![], rules: None };
+
+			// Check that the role does not exists already in the cluster
+			let roles_in_cluster = <Clusters<T>>::get(&cluster_id)
```
sinkingsugar: I'm sure that this can be done just using iter() without having to count vector members (and creating a vector..)

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+			{
+				return Err(Error::<T>::InvalidInput.into())
+			}
+
+			// Check that the caller is the owner of the cluster
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::SystematicFailure)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			// Check that the role exists in the cluster and in storage
+			let roles_in_cluster = <Clusters<T>>::get(&cluster_id)
```
sinkingsugar: See above, not really the way to do it

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+			ensure!(!role_name.len().is_zero(), Error::<T>::InvalidInput);
+
+			// only the owner of the cluster can do this operation
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			let role_name_vec = role_name.into_inner();
+
+			// Check that the role exists in the cluster and in storage
+			let roles_in_cluster = cluster
```
sinkingsugar: Above

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+			ensure!(!cluster.members.contains(&member), Error::<T>::MemberExists);
+
+			ensure!(cluster.members.len() < T::MembersLimit::get() as usize, Error::<T>::MembersLimitReached);
+
+			// Check that the roles for the member already exists in the cluster
+			let roles_in_cluster: Vec<Vec<u8>> =
+				cluster.roles.iter().map(|role| role.name.clone()).collect();
+
+			let roles = roles.clone();
+			for role in &roles {
```
sinkingsugar, line 423: Why clone if we use ref?

sinkingsugar (changes requested):
More issues. This PR has been a rust struggle 🙃

pallets/clusters/src/lib.rs
alebaffa: Done!

pallets/clusters/src/lib.rs
alebaffa: Done.

pallets/clusters/src/lib.rs
alebaffa: Done.

pallets/clusters/src/lib.rs
alebaffa: Oh right, this was something I forgot to remove after your previous review. Thanks!

pallets/clusters/src/lib.rs
sinkingsugar: There are way more places with this then I bothered commenting btw 😅

sinkingsugar: Also it's still wrong I think

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+			// write
+			// create an account for the cluster
+			let vault = get_vault_id(cluster_id);
+			<pallet_balances::Pallet<T> as fungible::Mutate<T::AccountId>>::mint_into(
+				&vault,
+				minimum_balance,
+			)?;
+
+			// create a pure proxy
+			pallet_proxy::Pallet::<T>::create_pure(
```
sinkingsugar: Did I use this on the other pallet? I don't remember this

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			// At creation there are no Members and no Rules assigned to a Role
+			let new_role = Role { name: role_name.into_inner(), members: vec![], rules: None };
+
+			// Check that the role does not exists already in the cluster
+			if <Clusters<T>>::get(&cluster_id)
+				.iter()
+				.any(|cluster| cluster.roles.iter().any(|role| new_role.name.eq(&role.name)))
```
sinkingsugar: I don't think the first iter any is correct, doesn't get return just a single cluster?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+			}
+
+			// Check that the caller is the owner of the cluster
+			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::SystematicFailure)?;
+			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+
+			// Check that the role exists in the cluster and in storage
+			if !<Clusters<T>>::get(&cluster_id)
+				.iter()
+				.any(|cluster| cluster.roles.iter().any(|role| role_name.eq(&role.name)))
```
sinkingsugar: like mentioned above

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+			{
+				return Err(Error::<T>::RoleNotFound.into())
+			}
+
+			let role_hash = blake2_128(&[&cluster_id[..], &role_name.as_slice()].concat());
+			let role_name_vec = role_name.into_inner();
+
+			// write
+			<Clusters<T>>::mutate(&cluster_id, |cluster| {
+				let cluster = cluster.as_mut().unwrap();
```
sinkingsugar: unwrap?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+				let index = cluster.roles.iter().position(|x| role_name_vec.eq(&x.name));
```
sinkingsugar: What's this?

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
sinkingsugar: Hmm I kinda see but I wonder if there is a better way

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+				// Add member into the cluster
+				cluster.members.push(member.clone());
+
+				// Associate the member with its roles in the cluster
+				for role in roles {
+					let index = cluster.roles.iter().position(|x| x.name == role);
+					if let Some(index) = index {
+						let role =
+							cluster.roles.get(index).ok_or(Error::<T>::SystematicFailure).unwrap();
+						let mut role_members = role.clone().members;
```
sinkingsugar: You are editing a clone of role also.. this likely won't work

pallets/clusters/src/lib.rs
alebaffa: At this point I am sure that this unwrap will not cause any problem. So it is safe to use it. We do this multiple times in all the other pallets too.

sinkingsugar: No that is not the right attitude honestly in such calls.
At least `except` but I prefer we handle properly, every `unwrap` is tech debt.

pallets/clusters/src/lib.rs
alebaffa: Oh my, yes of course. I have to stop coding at night, this is so stupid..

pallets/clusters/src/lib.rs
```diff
@@ -233,13 +234,22 @@ pub mod pallet {
...
 
-			// create a pure proxy
-			pallet_proxy::Pallet::<T>::create_pure(
-				origin.clone(),
-				T::ProxyType::default(),
-				T::BlockNumber::zero(),
-				0,
-			)?;
+			// create a proxy
+			let proxy_def = pallet_proxy::ProxyDefinition {
```
sinkingsugar, line 240: I mean if create_pure did the same thing I'm fine to use create_pure.... btw. Don't misunderstand my comment.

pallets/clusters/src/lib.rs
```diff
@@ -269,23 +279,21 @@ pub mod pallet {
 			let role_hash = blake2_128(&[&cluster_id.as_slice(), role_name.as_slice()].concat());
 
 			// Check that the caller is the owner of the cluster
-			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?;
-			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+			ensure!(who == <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?.owner, Error::<T>::NoPermission);
```
sinkingsugar: why this refactor? you are using `cluster` also under, so getting out once would have helped

pallets/clusters/src/lib.rs
```diff
@@ -269,23 +279,21 @@ pub mod pallet {
...
 
 			// At creation there are no Members and no Rules assigned to a Role
 			let new_role = Role { name: role_name.into_inner(), members: vec![], rules: None };
 
 			// Check that the role does not exists already in the cluster
-			if <Clusters<T>>::get(&cluster_id)
-				.iter()
-				.any(|cluster| cluster.roles.iter().any(|role| new_role.name.eq(&role.name)))
+			if <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?
```
sinkingsugar: optimize single get

pallets/clusters/src/lib.rs
```diff
@@ -269,23 +279,21 @@ pub mod pallet {
...
+				.roles.iter().any(|role| new_role.name.eq(&role.name))
 			{
 				return Err(Error::<T>::RoleExists.into())
 			}
 
 			// write
 			<Clusters<T>>::mutate(&cluster_id, |cluster| {
-				let cluster = cluster.as_mut().unwrap();
+				let cluster = cluster.as_mut().ok_or(Error::<T>::SystematicFailure).expect("Should find the cluster");
```
sinkingsugar: ok_or(_) + expect seems off. Pick one.

pallets/clusters/src/lib.rs
```diff
@@ -316,14 +324,12 @@ pub mod pallet {
 				return Err(Error::<T>::InvalidInput.into())
 			}
 
-			// Check that the caller is the owner of the cluster
-			let cluster = <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::SystematicFailure)?;
-			ensure!(who == cluster.owner, Error::<T>::NoPermission);
+			// Check that the caller is the owner of the cluster;
+			ensure!(who == <Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::SystematicFailure)?.owner, Error::<T>::NoPermission);
```
sinkingsugar: optimize, not sure why this was changed

pallets/clusters/src/lib.rs
```diff
@@ -316,14 +324,12 @@ pub mod pallet {
...
 
 			// Check that the role exists in the cluster and in storage
-			if !<Clusters<T>>::get(&cluster_id)
-				.iter()
-				.any(|cluster| cluster.roles.iter().any(|role| role_name.eq(&role.name)))
+			if !<Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?
```
sinkingsugar: oprimize single get

pallets/clusters/src/lib.rs
```diff
@@ -333,18 +339,23 @@ pub mod pallet {
...
-				}
+					 	for member in new_members_list {
+					 		members.push(member);
+					 	}
+				});
 			});
+
 			<RoleToSettings<T>>::mutate(&role_hash, |role| {
-				let role_settings = role.as_mut().unwrap();
+				let role_settings = role.as_mut().ok_or(Error::<T>::SystematicFailure).expect("Should find the role settings");
```
sinkingsugar: undecided error... return error or ensure or so

pallets/clusters/src/lib.rs
```diff
@@ -372,21 +383,21 @@ pub mod pallet {
 			let role_name_vec = role_name.into_inner();
 
 			// Check that the role exists in the cluster and in storage
-			if !<Clusters<T>>::get(&cluster_id)
-				.iter()
-				.any(|cluster| cluster.roles.iter().any(|role| role_name_vec.eq(&role.name)))
+			if !<Clusters<T>>::get(&cluster_id).ok_or(Error::<T>::ClusterNotFound)?
```
sinkingsugar: optimize get

pallets/clusters/src/lib.rs
```diff
@@ -372,21 +383,21 @@ pub mod pallet {
...
 
 			// write
 			// Remove Role from Cluster
 			<Clusters<T>>::mutate(&cluster_id, |cluster| {
-				let cluster = cluster.as_mut().unwrap();
-				let index = cluster.roles.iter().position(|x| role_name_vec.eq(&x.name));
-				if let Some(index) = index {
-					cluster.roles.remove(index);
-				}
+				let cluster = cluster.as_mut().ok_or(Error::<T>::SystematicFailure).expect("Should find the cluster");
```
sinkingsugar: ok_or + expect...

pallets/clusters/src/lib.rs
```diff
@@ -430,20 +441,19 @@ pub mod pallet {
 
 			// write
 			<Clusters<T>>::mutate(&cluster_id, |cluster| {
-				let cluster = cluster.as_mut().unwrap();
+				let cluster = cluster.as_mut().ok_or(Error::<T>::SystematicFailure).expect("Should find the cluster");
```
sinkingsugar: ok_or + expect

pallets/clusters/src/lib.rs
```diff
@@ -466,11 +476,13 @@ pub mod pallet {
 			// write
 			// Delete member from Cluster
 			<Clusters<T>>::mutate(&cluster_id, |cluster| {
-				let cluster = cluster.as_mut().unwrap();
-				let index = cluster.members.iter().position(|x| x == &member);
-				if let Some(index) = index {
-					cluster.members.remove(index);
-				}
+				let cluster = cluster.as_mut().ok_or(Error::<T>::SystematicFailure).expect("Should find the cluster");
```
sinkingsugar: ok_or + expect

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
alebaffa: I used both here because `ok_or` returns a `Result<T, E>`, which cannot be used in a closure that returns `()`, like in this case., I normally do `ok_or(..)?`, but since here I need the `ok` value of `Result`, I thought to use `expect` .
Would this work for you?

pallets/clusters/src/lib.rs
alebaffa: There is a detail that make it different, and it's the reason why I ended up copying from pallet_account, which I did not notice at first. It is in `pallet_proxy::Proxies::<T>::insert`.
The `create_pure` function internally creates a pure account that `insert `, while we are actually doing `insert `of an account that we create (in this case the `vault `- i.e. the account of the cluster).

pallets/clusters/src/lib.rs
alebaffa: Right..

pallets/clusters/src/lib.rs
alebaffa: Right...

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+			let role_hash = blake2_128(&[&cluster_id[..], &role_name.as_slice()].concat());
+			let role_name_vec = role_name.into_inner();
+
+			// write
+			<Clusters<T>>::mutate(&cluster_id, |cluster| {
+				let cluster = cluster
+					.as_mut()
+					.expect("Should find the cluster");
+				cluster.roles.iter().position(|x| x.name == role_name_vec).map(|index| {
+					let role = cluster
```
sinkingsugar: you are already iterating O(n) here https://marketplace.fragnova.com/proto-details/0x79543750bd76e74245ee62f3d92a77ba53cf49d934b1cf7196a242edf2138932
Take the index as well there.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+					for member in new_members_list {
+						members.push(member);
+					}
+				});
+			});
+
+			<RoleToSettings<T>>::mutate(&role_hash, |role| {
+				let role_settings = role
+					.as_mut()
+					.expect("Should find the role settings");
```
sinkingsugar: This could potentially fail, I don't see you fetching role_hash before in a safe manner

sinkingsugar: In fact what if does not exist?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+			}
+
+			// write
+			// Remove Role from Cluster
+			<Clusters<T>>::mutate(&cluster_id, |cluster| {
+				let cluster = cluster
+					.as_mut()
+					.expect("Should find the cluster");
+				cluster
+					.roles
```
sinkingsugar: optimize, avoid this search

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+				let cluster = cluster
+					.as_mut()
+					.expect("Should find the cluster");
+				// Add member into the cluster
+				cluster.members.push(member.clone());
+
+				// Associate the member with its roles in the cluster
+				for role in roles {
+					cluster
+						.roles
```
sinkingsugar, line 436: optimize

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+
+			// write
+			// Delete member from Cluster
+			<Clusters<T>>::mutate(&cluster_id, |cluster| {
+				let cluster = cluster
+					.as_mut()
+					.expect("Should find the cluster");
+
+				cluster
+					.members
```
sinkingsugar, line 467: optimize, contains iterates...

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
alebaffa: `role_hash` is generated from `cluster_id` and `role_name`. Both of them are checked before.
Actually, cluster_id implicitly, if a wrong one is passed and its related cluster is not found, it fails
```rust
get(&cluster_id).ok_or(Error::<T>::SystematicFailure)?;
```
and `role_name`
```rust
ensure!(!role_name.len().is_zero(), Error::<T>::InvalidInput);
```
So, if these two checks pass, a hash is surely generated and the mutate will not fail because of it.

alebaffa: The second question.. you are right. I need to add a check for the existence of the hash in the storage that I want to mutate.
I will add it.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+			// Check that the cluster does not already contain the member
+			ensure!(!cluster.members.contains(&member), Error::<T>::MemberExists);
+
+			ensure!(
+				cluster.members.len() < T::MembersLimit::get() as usize,
+				Error::<T>::MembersLimitReached
+			);
+
+			// Check that the roles for the member already exists in the cluster
+			let roles_in_cluster: Vec<Vec<u8>> =
```
aang114: we don't need to clone here. We can just borrow

pallets/clusters/src/lib.rs
aang114: How so 🤔? The only solution I see is having another storage map that maps cluster-id to roles (but we'll be sacrificing on storage)?

pallets/clusters/src/lib.rs
alebaffa: I can add a storage that maps a `cluster_id` to a `Vec<Role>` so I could avoid the iteration. I had a similar structure in a [previous commit](https://github.com/fragcolor-xyz/clamor/pull/224/commits/75884cf583d3db70a38355f2c0edebbfe1fb7e5e#diff-6804b4445503eaa0a518d6cabdb56e2901be2c3b0c19ba2b7b4a5e4becdad173), but it was mapping a `cluster_id` to a `Vec<role_hash>` .
Would that work (map(`cluster_id`, `Vec<Role>`)?

pallets/clusters/src/lib.rs
aang114: Although actually @alebaffa - you can make `cluster.roles` a `BTreeMap` (instead of a `Vec`) that maps a role to the list of members 🤔.  cc @sinkingsugar

pallets/clusters/src/lib.rs
sinkingsugar: Forget adding storage.
My point is simpler.
You are doing `cluster.roles.iter()` twice no? (including role_name check) so you could have taken the index before already no?

pallets/clusters/src/lib.rs
sinkingsugar: again role name index could have been taken before

pallets/clusters/src/lib.rs
sinkingsugar: why? to just ensure `ensure!(roles_in_cluster.contains(&role), Error::<T>::RoleNotFound);` you don't need this

pallets/clusters/src/lib.rs
aang114: Could be done without `iter()`, but not sure how one can reduce the time complexity 🤔

aang114 (commented):
LGTM, but it can be further optimised

pallets/clusters/src/lib.rs
alebaffa: Not sure to understand. It's not doing `cluster.roles.iter()` twice. The first `iter()` is for taking the index of the element in the vector via `position` and then i take the vector `cluster.roles` and i access to the element i need by its index (O(1)), without doing any other loop over the vector.

sinkingsugar: zoom out, way above you are doing exactly the same iteration.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+			ensure!(
+				who == cluster.owner,
+				Error::<T>::NoPermission
+			);
+
+			// Check that the role exists in the cluster and in storage
+			if !cluster
+				.roles
+				.iter()
+				.any(|role| role_name.eq(&role.name))
```
sinkingsugar: combine this with the one you will do in mutate to find index to avoid potentially 2 iterations while running extrinsic is the idea

sinkingsugar: but @aang114 suggestion is also good, BTMap do wonders

pallets/clusters/src/lib.rs
sinkingsugar: This one might be harder to oprimize true, but BTMap might help

alebaffa: I thought about BTMap, but discarded since the order of the elements is not important in this case and search is slower than Vec (O(log(n)) VS O(1)).
But i will look into it.

alebaffa: @aang114 if `cluster.roles` becomes a `BTreeMap<role.name, Vec<AccountId>>` then the `Role` struct needs to be stored somewhere else. I cannot use `Role` as key like `BTreeMap<Role, Vec<AccountId>> `. So, this option implies the creation of a `StorageMap<cluster_Id, Role>` (solutions I implemented in a [previous commit](https://github.com/fragcolor-xyz/clamor/pull/224/commits/75884cf583d3db70a38355f2c0edebbfe1fb7e5e#diff-6804b4445503eaa0a518d6cabdb56e2901be2c3b0c19ba2b7b4a5e4becdad173) btw, then discarded).

aang114: What about we change Role to:
```rust
struct Role<AccountId> {
   pub members: Vec<AccountId>,
   pub rules: Option<Rule>,
}
```

while `cluster.roles` is: `BTreeMap<role.name, Role<AccountId>>`. 

Does that work?

alebaffa: Same problem, `Role` is not stored anywhere. At this moment `Role` has not its own storage, it is included in `Clusters` storage itself (being part of `Cluster` struct as `Vec<Role>`).

alebaffa: A solution could be to have 
```rust
cluster.roles: BTreeMap<role.name, Role>
cluster.members: BTreeMap<AccountId, role.name>
```
It would make the job, but it feels weird.

I already thought about all this at the beginning of this implementation for `pallet-clusters`, I created storage like `ClusterToRole<role_hash, Role>`, but then I remembered that for our chain the priority should be put on reducing the storage rather than having a super optimized computational complexity, and so I removed them and ended up with this code.
Though, in `pallet-fragments` and `pallet-protos` we also have storage that has no functional meaning, but they allow to avoid iterations (such as fragments.`DataHashMap`, fragments.`DataHashMapIndex` or protos.`MetaKeys`, protos.`MetaKeysIndex`).
So, where is the boundary?

aang114: Yup, I know 🙂. I'm suggesting that we should replace `Vec<Role>` to a `BTreeMap<role.name, Role<AccountId>>` ? (where `Role` is now defined in the comment above)

alebaffa: Let me try with that thanks.

alebaffa: It would be interesting to discuss a bit deeper about the tradeoff storage VS compexity. In pallet-clusters I do not think that the numbers will be very high (i.e. a user unlikely will have thousands of clusters, and so a single cluster is unlikely to have thousands of members). But instead, in pallet-protos and pallet-fragments, that storage mentioned before is very likely to skyrocket pretty quickly since they are just indexes.

sinkingsugar: > Though, in pallet-fragments and pallet-protos we also have storage that has no functional meaning, but they allow to avoid iterations (such as fragments.DataHashMap, fragments.DataHashMapIndex or protos.MetaKeys, protos.MetaKeysIndex).
So, where is the boundary?

This is a good question, we have them for a few reasons:
1. To facilitate RPCs but not too much
2. To facilitate extrinsic functionality.

The issue here is that there will be a little amount of roles so simply it's not worth wasting time. As opposed to fragments.
I think this can stay as is even, all I want is best performance possible and correct code without any inefficient path.

alebaffa: sure, I understand that. OK, I see your points.

pallets/clusters/src/lib.rs
sinkingsugar: I'm wrong, it's multiple roles

sinkingsugar:
```suggestion
			let roles_in_cluster = cluster.roles.iter().map(|role| &role.name).collect::<Vec<&Vec<u8>>>();
```

sinkingsugar: @aang114 is right don't need clone.

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,520 @@
...
+			<Clusters<T>>::mutate(&cluster_id, |cluster| {
+				let cluster = cluster
+					.as_mut()
+					.expect("Should find the cluster");
+				cluster.roles.iter().position(|x| x.name == role_name_vec).map(|index| {
+					let role = cluster
+						.roles
+						.get(index)
+						.expect("Should find the cluster");
+					let mut members = role.clone().members;
```
sinkingsugar: This doesn't work... your clone is a clone... not the actual storage

sinkingsugar: So members is ephemeral

sinkingsugar:
> LGTM, but it can be further optimised

Let's remember that database migrations are to be avoided. Are we at least happy with storage layout?

pallets/protos/src/lib.rs
```diff
@@ -392,6 +395,7 @@ pub mod pallet {
 			tags: Vec<Vec<u8>>,
 			linked_asset: Option<LinkedAsset>,
 			license: UsageLicense<T::AccountId>,
+			cluster: Option<Cluster<T::AccountId>>,
```
sinkingsugar, line 398: Open continuation issue to implement proto side

sinkingsugar (approved):

pallets/protos/src/lib.rs
alebaffa: #239

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+				cluster_id,
+				roles: vec![],
+				members: vec![],
+			};
+
+			let minimum_balance =
+				<pallet_balances::Pallet<T> as fungible::Inspect<T::AccountId>>::minimum_balance();
+			let origin_balance =
+				<pallet_balances::Pallet<T> as fungible::Inspect<T::AccountId>>::balance(&who);
+			ensure!(origin_balance > minimum_balance, Error::<T>::InsufficientFunds);
```
aang114: why do we need to ensure that the user's balance is greater than the existential balance?

pallets/clusters/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+				delay: T::BlockNumber::default(),
+			};
+
+			let bounded_proxies: BoundedVec<_, T::MaxProxies> =
+				vec![proxy_def].try_into().map_err(|_| Error::<T>::TooManyProxies)?;
+
+			// ! Writing state
+
+			let deposit = T::ProxyDepositBase::get() + T::ProxyDepositFactor::get();
+			<T as pallet_proxy::Config>::Currency::reserve(&who, deposit)?;
```
aang114, line 252: I notice the sender reserves `desposit` over here 🤔 .

Okay nevermind I see why we need it to exist lol

pallets/clusters/src/lib.rs
aang114: nvm I see we do `reseve()` later. Could we change `>` to `≥` then 😅 ?

aang114 (approved):
LGTM. But why not change `>` to `≥` (see code comment made above) btw?...very minor lol

pallets/clusters/src/lib.rs
alebaffa: it is not needed, indeed. I can remove it.

PR #232, opened 2022-11-25
Monthly substrate bump

alebaffa:
Bump to fragnova-v0.9.33-1 (i.e. polkadot-v0.9.33).

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -48,14 +48,12 @@ where
 		+ Send
 		+ 'static,
 	C::Api: substrate_frame_rpc_system::AccountNonceApi<Block, AccountId, Index>,
-	C::Api: pallet_contracts_rpc::ContractsRuntimeApi<Block, AccountId, Balance, BlockNumber, Hash>,
 	C::Api: pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>,
 	C::Api: pallet_protos_rpc::ProtosRuntimeApi<Block, AccountId>,
 	C::Api: pallet_fragments_rpc::FragmentsRuntimeApi<Block, AccountId>,
 	C::Api: BlockBuilder<Block>,
 	P: TransactionPool + 'static,
 {
-	use pallet_contracts_rpc::{Contracts, ContractsApiServer};
 	use pallet_fragments_rpc::{FragmentsRpcServer, FragmentsRpcServerImpl};
 	use pallet_protos_rpc::{ProtosRpcServer, ProtosRpcServerImpl};
 	use pallet_transaction_payment_rpc::{TransactionPayment, TransactionPaymentApiServer};
@@ -69,7 +67,6 @@ where
 	// Making synchronous calls in light client freezes the browser currently,
 	// more context: https://github.com/paritytech/substrate/pull/3480
 	// These RPCs should use an asynchronous caller instead.
-	io.merge(Contracts::new(client.clone()).into_rpc())?;
 	io.merge(TransactionPayment::new(client.clone()).into_rpc())?;
 
 	io.merge(Dev::new(client.clone(), deny_unsafe).into_rpc())?;
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -146,6 +146,7 @@ pub fn new_partial(
 			registry: config.prometheus_registry(),
 			check_for_equivocation: Default::default(),
 			telemetry: telemetry.as_ref().map(|x| x.handle()),
+			compatibility_mode: Default::default(),
 		})?;
 
 	Ok(sc_service::PartialComponents {
@@ -302,6 +303,7 @@ pub fn new_full(
 				block_proposal_slot_portion: SlotProportion::new(2f32 / 3f32),
 				max_block_proposal_slot_portion: None,
 				telemetry: telemetry.as_ref().map(|x| x.handle()),
+				compatibility_mode: Default::default(),
 			},
 		)?;
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Check if all the Polkadot.js RPC tests pass
         run: |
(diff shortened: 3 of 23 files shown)


alebaffa:
Biggest changes:
- the removal of RPC from pallet-contracts that has impacts in our pallet-protos where we call [bare_call](https://github.com/fragcolor-xyz/clamor/blob/db8245faadb1dfd24688ae4e9a47164dc444743e/pallets/protos/src/lib.rs#L965)
- change in the runtime, notably how [block weight](https://github.com/fragcolor-xyz/clamor/blob/db8245faadb1dfd24688ae4e9a47164dc444743e/runtime/src/lib.rs#L213) is calculated. This update follows the implementation of the [template node](https://github.com/paritytech/substrate/blob/polkadot-v0.9.33/bin/node-template/runtime/src/lib.rs). Basically, the code has been refactored to extract the calculations for extrinsic types.

node/src/rpc.rs
```diff
@@ -69,7 +67,6 @@ where
 	// Making synchronous calls in light client freezes the browser currently,
 	// more context: https://github.com/paritytech/substrate/pull/3480
 	// These RPCs should use an asynchronous caller instead.
-	io.merge(Contracts::new(client.clone()).into_rpc())?;
```
sinkingsugar, line 72: I find this weird. How are reads supposed to happen now?

sinkingsugar (changes requested):

node/src/rpc.rs
alebaffa: So, there has been a change in the general management of RPC in Substrate to instead make `state_call`. This is mentioned in this issue  [Support for runtime api calls (via state_call)](https://github.com/paritytech/capi/discussions/218) where there's a mention about the history of this decision:
> I was talking about the general direction in Substrate to replace RPC calls with runtime calls. For the end user of a JS library it doesn't make much difference what is used under the hood.
Some RPCs will be removed from pallets, as they are only channeling data to state_call

A related issue  '[RPC Calls vs. state_calls](https://github.com/paritytech/capi/issues/228)' the technical reasons are explained more in detail:
> What you want to do is use state_call whenever possible and only fall back to a custom RPC where no runtime API exists (storage access would be one example). Many of the RPCs are just wrappers around a runtime API and are redundant. I want to remove them.

This has actually started in June with this issue in polkadot-js '[Convenient way to invoke runtime API](https://github.com/polkadot-js/api/issues/4930)', 
>We would like to replace custom RPC with state_call but we need a better API to invoke runtime APIs.
We should be able to supply a definition (similar to RPC defs) and polkadot.js should be able to use it to generate methods like api.call.evm.execute(contract_addr, input) and return the decoded response.
In future, if we can make metadata to include runtime API defs, then we can automatically generate those methods from metadata.

The discussion went on in July with this issue in substrate-connect '[Custom getters](https://github.com/paritytech/substrate-connect/issues/1096)' that triggered another issue in August '[Replace contracts RPC calls with state_call](https://github.com/paritytech/cargo-contract/issues/683)' where all officially started in Substrate:
> The contract specific **RPC endpoints are planned to be removed**. These should be replaced by calling the `state_call` RPC with SCALE encoded calls for the various contracts extrinsics,

So, since Substrate pallets are going to move from RPC calls to `state_calls`, `pallet-contracts` removed them with [this PR](https://github.com/paritytech/substrate/pull/12358) in Substrate:
> Those RPCs are redundant and should be called via state_call. Polkadot.js and cargo contract both already migrated. This means all our tooling should not break after this change.
> The high level contracts API in polkadot.js already uses state_call so upgrading to the latest version here is enough. If you are using api.rpc.contracts directly you need to switch to api.call.contractsApi.

In PolkadotJS the rpc calls have been indeed [deprecated](https://github.com/polkadot-js/api/commit/a92bfca548b0a7620de9d79a2e3a50f68fefe2a8).

sinkingsugar: how does this impact our rpcs?

node/src/service.rs
```diff
@@ -146,6 +146,7 @@ pub fn new_partial(
 			registry: config.prometheus_registry(),
 			check_for_equivocation: Default::default(),
 			telemetry: telemetry.as_ref().map(|x| x.handle()),
+			compatibility_mode: Default::default(),
```
sinkingsugar, line 149: whats this?

sinkingsugar (approved):

node/src/service.rs
alebaffa: It's due to [this change](https://github.com/paritytech/substrate/commit/4e1e17cccd499dfe49e8c1bed01957953aa4c839) in Aura.

node/src/rpc.rs
alebaffa: @sinkingsugar 
So, there are no impacts for our rpcs. Leaving them as they are will not create problems.

The change they started to implement (and push people to as well) is to be able to use runtime APIs calls directly because 
> they are on-chain upgradable (unlike rpc calls) ([source](https://github.com/paritytech/substrate/issues/11648#issuecomment-1191457418))

The next step seems to add them into the metadata of the chain.

I am still trying to understand what this would mean on frontend side (js) in case we decide to adopt the same approach. I am reading all the issues related to this part, notably [here](https://github.com/polkadot-js/api/issues/5083) and [here](https://github.com/polkadot-js/api/pull/4987). 

My understanding so far is that this would allow us (on clamor side) to 
- get rid of `rpc/runtime-api` in protos and fragments that include their `pub trait ProtosRuntimeApi` and `FragmentsRuntimeApi`
- add those traits under `sp_api::decl_runtime_apis!` directly into `[protos|fragments]/src/lib.rs`
- add `impl pallet_protos::ProtosApi` directly into `runtime/src/lib.rs`

On the frontend side, instead of doing 
```javascript
  const api = await ApiPromise.create({
    provider: wsProvider,
    rpc: {
      protos: {
```
it could do
```javascript
  const api = await ApiPromise.create({
    provider: wsProvider,
    runtime: {
      ProtosApi: [
        {
```

I can create another issue dedicated to this to continue further with the investigation for this.

sinkingsugar: I think we need to adopt it because potentially removes code to maintain, but we do some transformations so not sure it's possible

sinkingsugar (approved):

PR #236, opened 2022-12-07
Rollback to substrate tag fragnova-v0.0.6

alebaffa:
Rollback needed after IPFS issues [introduced](https://github.com/fragcolor-xyz/substrate/commit/50325f1f0b36210d1730f7c903ab89083fe37a05) with Substrate tag `fragnova-v0.29.1`.

Rolled back to this tag where IPFS works with `go-ipfs_v0.12.0`.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -16,10 +16,10 @@
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use frame_support::{
-	traits::{ConstU128, ConstU32, ConstU64},
-	dispatch::DispatchClass,
+	traits::{ConstU16, ConstU128, ConstU32, ConstU64},
+	weights::DispatchClass,
 };
-use frame_system::{EnsureRoot, EnsureSigned};
+use frame_system::{EnsureRoot, EnsureSigned, limits::{BlockLength, BlockWeights},};
 use pallet_grandpa::{
 	fg_primitives, AuthorityId as GrandpaId, AuthorityList as GrandpaAuthorityList,
 };
@@ -93,10 +93,10 @@ pub type Hash = sp_core::H256;
 
 /// Unchecked extrinsic type as expected by this runtime.
 pub type UncheckedExtrinsic =
-	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
+	generic::UncheckedExtrinsic<Address, Call, Signature, SignedExtra>;
 
 /// The payload being signed in transactions.
-pub type SignedPayload = generic::SignedPayload<RuntimeCall, SignedExtra>;
+pub type SignedPayload = generic::SignedPayload<Call, SignedExtra>;
 
 /// Related to Index pallet
 pub type AccountIndex = u64;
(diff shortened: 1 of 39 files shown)


alebaffa:
All functionalities introduced in our pallets since then remained untouched, obviously.

sinkingsugar:
Not sure I get it, is this a backport?
New devel functionalities are IN or not?
Also does IPFS test work fully?

alebaffa:
I moved back to `fragnova-v0.0.6` tag, so keeping all latest functionalities from `devel`. 
IPFS works fully, I tested manually with go-ipfs_v0.12.0.

sinkingsugar:
> I moved back to `fragnova-v0.0.6` tag, so keeping all latest functionalities from `devel`. IPFS works fully, I tested manually with go-ipfs_v0.12.0.

Nice, are the tests also fixed? the ipfs test

shards/test-ipfs.sh
```diff
@@ -4,9 +4,9 @@ set -e
 # INSTALL IPFS
 apt-get update
 apt-get install -y wget
-wget -q https://dist.ipfs.tech/kubo/v0.16.0/kubo_v0.16.0_linux-amd64.tar.gz
-tar -xvzf kubo_v0.16.0_linux-amd64.tar.gz
-export PATH=$PATH:`pwd`/kubo
+wget -q https://dist.ipfs.tech/go-ipfs/v0.12.0/go-ipfs_v0.12.0_linux-amd64.tar.gz
```
sinkingsugar, line 7: kubo 0.17, let's do a try with that too? or doesn't work?

runtime/src/lib.rs
```diff
@@ -210,13 +215,27 @@ parameter_types! {
...
+		.for_class(DispatchClass::Operational, |weights| {
+			weights.max_total = Some(MAXIMUM_BLOCK_WEIGHT);
+			// Operational transactions have some extra reserved space, so that they
+			// are included even if block reached `MAXIMUM_BLOCK_WEIGHT`.
+			weights.reserved = Some(
+				MAXIMUM_BLOCK_WEIGHT - NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT
+			);
+		})
+		.avg_block_initialization(AVERAGE_ON_INITIALIZE_RATIO)
+		.build_or_panic();
```
sinkingsugar, line 238: let's make sure this is correct

sinkingsugar (approved):
Approved with some comments!

shards/test-ipfs.sh
alebaffa: It doesn't work with kubo 0.17.

runtime/src/lib.rs
alebaffa: Yes, it is the version there was back then.
The [new version](https://github.com/paritytech/substrate/blob/ea3ca3f757ff9d9559665719a77da81f4cf0f0ce/bin/node-template/runtime/src/lib.rs#L138), used also in latest node-template, is a refactoring that extracts this calculation into `fn with_sensible_defaults` in [frame/system/limits.rs](https://github.com/paritytech/substrate/blob/ea3ca3f757ff9d9559665719a77da81f4cf0f0ce/frame/system/src/limits.rs#L323). In addition, the new version uses a utility function `set_proof_size`, which has been introduced in `weight_v2` even [after](https://github.com/paritytech/substrate/commit/17c07af0b953b84dbe89341294e98e586f9b4591) the substrate tag that introduced the problems with ipfs.

shards/test-ipfs.sh
sinkingsugar: That's so weird to be honest but must be libp2p incompatibilities somehow ( a library engineered for compatibility 😄 )

PR #238, opened 2022-12-11
Bounded all extrinsics' `Vec` params, if the `Vec` param is stored into state

aang114:
these are the bounding parameters:
- `pallet_protos::Config::StringLimit` for limiting string size (used by both `pallet_protos` and `pallet_fragments` - not sure if `pallet_fragments` should have its own Config @sinkingsugar @alebaffa 😅)
- `pallet_protos::Config::MaxTags` for limiting max number of proto tags
- `pallet_protos::Config::DetachAccountLimit` for limiting the maximum length that a detach account can be. (used by both `pallet_protos` and `pallet_fragments` - not sure if `pallet_fragments` should have its own Config @sinkingsugar @alebaffa 😅)

### DIFF
--- a/pallets/fragments/src/lib.rs
+++ b/pallets/fragments/src/lib.rs
@@ -543,6 +543,8 @@ pub mod pallet {
 		ProtoOwnerNotFound,
 		/// No Permission
 		NoPermission,
+		/// Detach Request's Target Account is empty
+		DetachAccountIsEmpty,
 		/// Detach Request Already Submitted
 		DetachRequestAlreadyExists,
 		/// Already detached
@@ -575,6 +577,10 @@ pub mod pallet {
 		UniqueDataExists,
 		/// Currency not found
 		CurrencyNotFound,
+		/// Fragment Definition's Metadata key is empty
+		DefinitionMetadataKeyIsEmpty,
+		/// Fragment Instance's Metadata key is empty
+		InstanceMetadataKeyIsEmpty,
 	}
 
 	// Dispatchable functions allows users to interact with the pallet and invoke state changes.
@@ -716,6 +722,8 @@ pub mod pallet {
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
+			ensure!(!metadata_key.is_empty(), Error::<T>::DefinitionMetadataKeyIsEmpty);
+
 			let proto_hash =
 				<Definitions<T>>::get(definition_hash).ok_or(Error::<T>::NotFound)?.proto_hash; // Get `proto_hash` from `definition_hash`
 			let proto: Proto<T::AccountId, T::BlockNumber> =
@@ -801,6 +809,8 @@ pub mod pallet {
 			data: Vec<u8>,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
+
+			ensure!(!metadata_key.is_empty(), Error::<T>::InstanceMetadataKeyIsEmpty);
 
 			ensure!(!<DetachedHashes<T>>::contains_key(&DetachHash::Instance(definition_hash, Compact(edition_id), Compact(copy_id))), Error::<T>::Detached);
 
@@ -1387,6 +1397,8 @@ pub mod pallet {
 		) -> DispatchResult {
 
 			let who = ensure_signed(origin)?;
+
+			ensure!(!target_account.is_empty(), Error::<T>::DetachAccountIsEmpty);
 
 			let current_block_number = <frame_system::Pallet<T>>::block_number();
 
--- a/pallets/fragments/src/lib.rs
+++ b/pallets/fragments/src/lib.rs
@@ -166,9 +166,9 @@ pub enum Currency<TFungibleAsset> {
 
 /// **Struct** of a **Fragment Definition's Metadata**
 #[derive(Encode, Decode, Clone, scale_info::TypeInfo, Debug, PartialEq)]
-pub struct DefinitionMetadata<TFungibleAsset> {
+pub struct DefinitionMetadata<TU8Vector, TFungibleAsset> {
 	/// **Name** of the **Fragment Definition**
-	pub name: Vec<u8>,
+	pub name: TU8Vector,
 	/// **Currency** that must be used to buy **any and all Fragment Instances created from the Fragment Definition**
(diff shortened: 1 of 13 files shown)


pallets/fragments/src/lib.rs
```diff
@@ -166,9 +166,9 @@ pub enum Currency<TFungibleAsset> {
 
 /// **Struct** of a **Fragment Definition's Metadata**
 #[derive(Encode, Decode, Clone, scale_info::TypeInfo, Debug, PartialEq)]
-pub struct DefinitionMetadata<TFungibleAsset> {
+pub struct DefinitionMetadata<TU8Vector, TFungibleAsset> {
```
sinkingsugar, line 169: Why?

pallets/fragments/src/lib.rs
```diff
@@ -183,11 +183,11 @@ pub struct UniqueOptions {
 
 /// **Struct** of a **Fragment Definition**
 #[derive(Encode, Decode, Clone, scale_info::TypeInfo, Debug, PartialEq)]
-pub struct FragmentDefinition<TFungibleAsset, TAccountId, TBlockNum> {
+pub struct FragmentDefinition<TU8Array, TFungibleAsset, TAccountId, TBlockNum> {
```
sinkingsugar, line 186: Also why?

pallets/fragments/src/lib.rs
```diff
@@ -342,7 +342,7 @@ pub mod pallet {
 		_,
 		Identity,
 		Hash128,
-		FragmentDefinition<T::AssetId, T::AccountId, T::BlockNumber>,
+		FragmentDefinition<Vec<u8>, T::AssetId, T::AccountId, T::BlockNumber>,
```
sinkingsugar, line 346: Not useful consider we use definition only here

pallets/fragments/src/lib.rs
```diff
@@ -1399,7 +1405,7 @@ pub mod pallet {
 			}
 
 			let detach_hash = DetachHash::Instance(definition_hash, Compact(edition_id), Compact(copy_id));
-			let detach_request = DetachRequest { hash: detach_hash.clone(), target_chain, target_account};
+			let detach_request = DetachRequest { hash: detach_hash.clone(), target_chain, target_account: target_account.into() };
```
sinkingsugar, line 1421: I suppose @alebaffa you do something similar iirc anyway already. @aang114 Alessandro iirc also ensures non zero! Which is a good touch in some cases.

sinkingsugar (changes requested):

pallets/fragments/src/lib.rs
alebaffa: You do not need to modify the structs fields to be `BoundedVec`. Only having them as inputs in the extrinsics, as you already did btw, is enough. It's just a check on the inputs from the caller, you can keep them as `Vec<u8>` for storage, so consequently in the structs.

pallets/fragments/src/lib.rs
alebaffa: I ensure non zero in pallet-oracle, [here](https://github.com/fragcolor-xyz/clamor/blob/458a5f20d20d805e768dbfcd15ea5b4d9c0041a2/pallets/oracle/src/lib.rs#L424):
```rust
ensure!(round_id.gt(&U256::zero()), "Price from oracle is 0");
ensure!(price.gt(&U256::zero()), "Price from oracle is <= 0");
```

sinkingsugar: err we are talking of bounded vectors though

pallets/fragments/src/lib.rs
aang114: > Why?

Because `fragments.create()` has a parameter of type `DefinitionMetadata`. Therefore, if we don't bound that parameter, a definition's name can be of any length!

pallets/fragments/src/lib.rs
aang114: > Also why?

reason give above (basically because `DefinitionMetadata` is a parameter of one of the extrinsics)

pallets/fragments/src/lib.rs
aang114: I bounded the parameter in `fragments.create()` rather than using `ensure!()` because then will even prevent the extrinsic from entering into the block if its `DefinitionMetadata` parameter is too long. so this prevents polluting/spamming of blocks

pallets/fragments/src/lib.rs
aang114: reason give above (basically because `DefinitionMetadata` is a parameter of one of the extrinsics)

pallets/fragments/src/lib.rs
aang114: > I suppose @alebaffa you do something similar iirc anyway already. @aang114 Alessandro iirc also ensures non zero! Which is a good touch in some cases.

done

pallets/protos/src/lib.rs
```diff
@@ -418,6 +420,8 @@ pub mod pallet {
 
 			ensure!(!data.is_empty(), Error::<T>::ProtoDataIsEmpty);
 
+			ensure!(!tags.iter().enumerate().any(|(index, tag)| tags.iter().enumerate().any(|(i, t)| t == tag && i != index)), Error::<T>::DuplicateProtoTagExists); // TODO Review - Is `O(n ^ 2)` good? (Alternatively we can **use HashMap** or **sort the tags then check for equal consecutive elements** -  but I don't think it's worth it since `T::MaxTags` is small
```
sinkingsugar, line 427: it's fine since t(max) is predictable

sinkingsugar (approved):

pallets/fragments/src/lib.rs
aang114: > Why?

so is it okay, @sinkingsugar ?

pallets/protos/src/lib.rs
aang114: is this reason okay? - https://github.com/fragcolor-xyz/clamor/pull/238#discussion_r1045316357 
@sinkingsugar

pallets/fragments/src/lib.rs
sinkingsugar: yeah ok

PR #242, opened 2022-12-20
Optimize clusters and refactor edit_role

alebaffa:
This PR optimizes the storage layer in `pallet-clusters`.

Another change in this PR is about `edit_role`, now replaced with smaller functions. I realized that it was somehow overlapping with `add_member` function. The former was adding a `Vec<AccountId>` into the list of members, while the latter was adding a single` AccountId` in the list of members, both of them assigning roles to the members. I found it confusing, so I split `edit_role` into the following smaller functions:
- `add_role_members`, `delete_role_members`, `add_role_settings`, `delete_role_settings`

This makes the whole code easier to read and also more intuitive from a user/API perspective.

### DIFF
--- a/pallets/clusters/src/lib.rs
+++ b/pallets/clusters/src/lib.rs
@@ -8,7 +8,7 @@ use codec::{Decode, Encode};
 use frame_support::BoundedVec;
 pub use pallet::*;
 use sp_clamor::Hash128;
-use sp_std::{vec, vec::Vec};
+use sp_std::{vec, vec::Vec, collections::btree_map::BTreeMap};
 
 #[cfg(test)]
 mod mock;
@@ -33,11 +33,9 @@ pub struct RoleSetting {
 
 /// **Struct** of **Role** belonging to a **Cluster**.
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
-pub struct Role<TAccountId> {
+pub struct Role {
 	/// Name of the role
 	pub name: Vec<u8>,
-	/// The list of Members associated with the role
-	pub members: Vec<TAccountId>,
 	/// The optional list of Rules associated to the Role
 	pub rules: Option<Rule>,
 	/// The settings of the Role
@@ -60,10 +58,8 @@ pub struct Cluster<TAccountId> {
 	pub name: Vec<u8>,
 	/// The ID of the cluster
 	pub cluster_id: Hash128,
-	/// The list of Roles associated to the Cluster
-	pub roles: Vec<Role<TAccountId>>,
-	/// The list of Members of the Cluster
-	pub members: Vec<TAccountId>,
+	/// The map that contains the list of Roles in Cluster, each associated with an index
+	pub roles: BTreeMap<u64, Role>,
 }
 
 /// **Struct** representing the details about accounts created off-chain by various parties and integrations.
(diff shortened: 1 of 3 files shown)


pallets/clusters/src/lib.rs
```diff
@@ -130,6 +126,19 @@ pub mod pallet {
...
 	pub type ClustersByOwner<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, Vec<Hash128>>;
 
+	/// **StorageDoubleMap** that maps a (**AccountId**, **Cluster hash**) with a list of **Role** indexes.
+	#[pallet::storage]
+	pub type Members<T: Config> =
+		StorageDoubleMap<_, Twox64Concat, T::AccountId, Identity, Hash128, Vec<u64>>;
+
+	/// **StorageMap** that maps a **(Cluster hash, role name)** to an **index number**
+	#[pallet::storage]
+	pub type ClusterRoleKeys<T: Config> = StorageMap<_, Twox64Concat, (Hash128, Vec<u8>), u64>;
```
sinkingsugar: Isn't this better to be a doublemap though? how is it used?

alebaffa: I did not use it because I never need to filter by one of the two keys, which is the main advantage of the `StorageDoubleMap`.

sinkingsugar: If you don't use it we don't need it then?

alebaffa: Actually, I could get rid of it, since the index is already included in the `Cluster.roles: BTreeMap<u64, Role>`.
Let me see.

alebaffa: No well, I cannot remove it. The idea is:

- Cluster contains its list of roles declared as `roles: BTreeMap<role_index, Role struct>`
- the map between the `Role` in a cluster and its `index` is maintained in `ClusterRoleKeys`, which is a `StorageMap` having a tuple` (cluster_hash, role_name)` as a key. I need a tuple because a role name is unique inside a cluster, but there can be roles with the same name in different clusters.  `ClusterRoleKeys` allows to access the `BTreeMap` mentioned before
- The role index is a counter kept in `ClusterRoleKeysIndex`, which is a `StorageValue`. The counter increases every time a new role is created or a new member (with a new role) is added into the cluster.


Basically, `ClusterRoleKeys` and `ClusterRoleKeysIndex` are just utility storages.

pallets/clusters/src/lib.rs
```diff
@@ -60,10 +58,8 @@ pub struct Cluster<TAccountId> {
 	pub name: Vec<u8>,
 	/// The ID of the cluster
 	pub cluster_id: Hash128,
-	/// The list of Roles associated to the Cluster
-	pub roles: Vec<Role<TAccountId>>,
-	/// The list of Members of the Cluster
-	pub members: Vec<TAccountId>,
+	/// The map that contains the list of Roles in Cluster, each associated with an index
+	pub roles: BTreeMap<u64, Role>,
```
sinkingsugar: index is confusing, that's actually the ID of the name no?

pallets/clusters/src/lib.rs
sinkingsugar: I'm a bit confused, I will check it better tomorrow but the idea here is to implement it like Tags in protos pallet. Is that the intention?
Basically we want to use some short IDs instead of long names and as well we want to store short IDs instead of full data structures nested.

sinkingsugar: How do I iterate all members of a cluster btw? for example

pallets/clusters/src/lib.rs
```diff
@@ -130,6 +126,19 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type ClustersByOwner<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, Vec<Hash128>>;
 
+	/// **StorageDoubleMap** that maps a (**AccountId**, **Cluster hash**) with a list of **Role** indexes.
+	#[pallet::storage]
+	pub type Members<T: Config> =
+		StorageDoubleMap<_, Twox64Concat, T::AccountId, Identity, Hash128, Vec<u64>>;
```
sinkingsugar: We cannot iterate all members of a cluster this way, should be reverse order to do a partial key query no?

pallets/clusters/src/lib.rs
sinkingsugar: Ok apart from the fact we need to iterate members of a cluster.

I think that the index counter should be inside the cluster; and could be a u16, we don't need more than u16::max roles most likely. Can be u32 to be safe even 🤔

The Account map could just contain clusters.
While the double storage should start with the cluster hash in order to iterate members. This setup goes full circle. As in we can go from account to cluster and deep inside and reverse from cluster to account.

sinkingsugar: Don't you agree that the most common thing users will do is check which are the clusters members? Btw

pallets/clusters/src/lib.rs
alebaffa: yeah, actually another index is not be needed at all, I realize. This could be a `BTreeMap<Vec<u8>, Role>` with `Vec<u8>` being simply the role name.

sinkingsugar: 👎 we want to cache names like tags in protos actually. We don't want repeating names. Check protos metadata. What I meant was more regarding the fact it was defined index in your comments, it's more an ID of the name instead (like in protos, if we called it index in protos it's also not correct btw)

sinkingsugar: Let's have a meeting about this. Let's avoid too much back and forth again.

sinkingsugar (commented):
It's good.
I just wonder if we really need this **add settings* rather than a simple **add single setting** (same for remove) which is way easier code.... (although it seems correct)

alebaffa:
I implemented it thinking what I would expect from such a functionality. Though it's true that `delete_member` and `delete_role` handle a single item, not a list of items.

sinkingsugar:
> I implemented it thinking what I would expect from such a functionality. Though it's true that `delete_member` and `delete_role` handle a single item, not a list of items.

Fair enough, I just find it harder to review and to predict what could go wrong :)

sinkingsugar (approved):
LGTM

PR #243, opened 2022-12-20
Allow a Collection of Proto-Fragments or a Collection of Fragment-Instances to be detached in single call

aang114:
Refactored `fragments.detach()` and `protos.detach()` to accept a list of Proto-Fragments and a list of Fragment-Instances (of a single non-copyable FD) respectively.

By calling any one of the two extrisnics - it will **ultimately cause an event will be emitted** which will **contain the merkle-root of the list of protos/FIs**


pallets/detach/src/lib.rs
```diff
@@ -486,7 +503,8 @@ pub mod pallet {
 					// We can still have multiple transactions compete for the same "spot",
 					// and the one with higher priority will replace other one in the pool.
 					.and_provides((
-						data.hash.clone(),
+						data.hashes.clone(), // TODO Review - Isn't it redundant to have `data.hashes` as part of the "provides" tag? Isn't `data.merkle_root` enough?
```
sinkingsugar: I think it's redundant for the reason you mention yes.

pallets/fragments/src/lib.rs
```diff
@@ -1383,45 +1384,63 @@ pub mod pallet {
...
 			ensure!(
-				owned_instances.contains(&(Compact(edition_id), Compact(copy_id))),
+				(definition_permissions & FragmentPerms::COPY) != FragmentPerms::COPY,
 				Error::<T>::NoPermission
 			);
 
-			// REVIEW - no go if will expire this block
-			let instance = <Fragments<T>>::get((definition_hash, edition_id, copy_id))
-				.ok_or(Error::<T>::NotFound)?;
-			if let Some(expiring_at) = instance.expiring_at {
```
sinkingsugar, line 1416: I think for now any item that can expire should not be detached at all

sinkingsugar (changes requested):
Great! I think we need more tests though.

pallets/fragments/src/lib.rs
```diff
@@ -1383,45 +1384,63 @@ pub mod pallet {
...
 			target_account: BoundedVec<u8, T::DetachAccountLimit>, // an eth address or so
 		) -> DispatchResult {
-
 			let who = ensure_signed(origin)?;
 
+			ensure!(!edition_ids.is_empty(), Error::<T>::InstancesToDetachIsEmpty);
 			ensure!(!target_account.is_empty(), Error::<T>::DetachAccountIsEmpty);
 
-			let current_block_number = <frame_system::Pallet<T>>::block_number();
+			// let current_block_number = <frame_system::Pallet<T>>::block_number();
```
sinkingsugar: We don't use this?

pallets/fragments/src/lib.rs
```diff
@@ -1383,45 +1384,63 @@ pub mod pallet {
...
+
+			let definition_permissions =
+				<Definitions<T>>::get(definition_hash).ok_or(Error::<T>::NotFound)?.permissions;
 
-			// owner must own instance
-			let owned_instances = <Inventory<T>>::get(who, definition_hash).ok_or(Error::<T>::NoPermission)?;
 			ensure!(
-				owned_instances.contains(&(Compact(edition_id), Compact(copy_id))),
+				(definition_permissions & FragmentPerms::COPY) != FragmentPerms::COPY,
```
sinkingsugar: btw also COPIABLE items cannot be detached I guess for now

pallets/fragments/src/lib.rs
aang114: But if a Fragment Definition is not copyable, it cannot expire. That's why I don't check

pallets/fragments/src/lib.rs
aang114: No we don't. Since we don't need to check for expiry since **if a Fragment Definition is not copyable, it cannot expire**

pallets/fragments/src/lib.rs
aang114: Yes. That's why I check if the Fragment Definition is copyable - rather than checking if the item is copyable

aang114: > btw also COPIABLE items cannot be detached I guess for now

Hmm but some copyable items can be `give()`en and its copy permission removed 🤔. That's true.

**So should I also accept Copy ID(s) as a parameter? ** Cos in the spec you previously mentioned, you only wanted to accept `edition_ids` @sinkingsugar

sinkingsugar: Did we clarify this on discord @aang114 ?

pallets/detach/src/lib.rs
```diff
@@ -486,7 +543,7 @@ pub mod pallet {
 					// We can still have multiple transactions compete for the same "spot",
 					// and the one with higher priority will replace other one in the pool.
 					.and_provides((
-						data.hash.clone(),
+						data.merkle_root, // TODO Review - what if by co-incidence two collections that contain different things have the same merkle-root?
```
alebaffa: With a 256 bit hash function, a merkle root collision is likely to be probabilistically impossible. In general, from my readings, with a hash function >= 64 bit you can be relatively safe.

sinkingsugar: To answer his comment.
Stars would collide and we would be sucked by a black hole if that happens.

sinkingsugar: It's blake2 no?

sinkingsugar (changes requested):

pallets/detach/src/uomo_contento.rs
```diff
@@ -0,0 +1,140 @@
+//! Everything in this crate was copied from: https://crates.io/crates/beefy-merkle-tree/10.0.0
```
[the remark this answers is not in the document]
alebaffa: oh thank you, seriously. I was about to write the same comment.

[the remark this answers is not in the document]
alebaffa: A note: always keep in mind that the code you write today will likely be seen/modified/inspected by someone else months/years from now. And it may also happen that you will not be here anymore to help that person. So, meaningful naming is important in a code base, as a knowledge sharing practice.

alebaffa (approved):

pallets/detach/src/uomo_contento.rs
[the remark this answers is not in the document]
aang114: > Please give a meaningful name to this file.

I named it this because I wanted to emphasize that this is only a temporary arrangement (hopefully)

pallets/detach/src/lib.rs
aang114: > It's blake2 no?

the merkle root is calculated using keccak256

aang114:
Hi @sinkingsugar, please approve this PR (if okay) so that I can merge it. Thanks

pallets/detach/src/lib.rs
sinkingsugar: that's ok too!

pallets/detach/src/copy_pasted_beefy_merkle_tree.rs
```diff
@@ -0,0 +1,144 @@
+//! Everything in this crate was copied from: https://crates.io/crates/beefy-merkle-tree/10.0.0 (which is also used in Substrate)
```
sinkingsugar, line 1: less silly name please.
Can you simply seriously call it `merkle_tree.rs`? You explained the rest in comments...

sinkingsugar (changes requested):

pallets/detach/src/copy_pasted_beefy_merkle_tree.rs
aang114: done

pallets/detach/src/lib.rs
```diff
@@ -85,22 +87,49 @@ use frame_system::offchain::{
...
+pub enum DetachCollection {
+	/// List of Proto-Fragments (identified by its Hash)
+	Protos(Vec<Hash256>),
+	/// List of Fragment Instances (identified as a tuple of its Fragment Definition Hash, its Edition ID and its Copy ID)
+	Instances(Vec<(Hash128, Compact<InstanceUnit>, Compact<InstanceUnit>)>),
+}
+impl DetachCollection {
+
+	/// Get the Collection type
+	fn get_type(&self) -> Vec<u8> {
```
sinkingsugar: What's the point of this?

pallets/detach/src/lib.rs
```diff
@@ -358,15 +393,37 @@ pub mod pallet {
...
+						log::debug!("Detached hash: {:?} signature: {:?}", detach_hash, data.remote_signature);
+					});
+
+				}
+			}
+
+			Self::deposit_event(Event::CollectionDetached {
+				merkle_root: data.merkle_root,
+				remote_signature: data.remote_signature,
+				collection_type: data.collection.get_type(),
```
sinkingsugar, line 434: enum or nothing here.

pallets/detach/src/lib.rs
sinkingsugar: remove comment btw

pallets/detach/src/lib.rs
```diff
@@ -575,14 +632,23 @@ pub mod pallet {
 			}
 		}
 
+		fn get_merkle_root(detach_collection: &DetachCollection) -> Hash256 {
+			let detach_hashes = detach_collection.get_abi_encoded_hashes();
+			// TODO Review - Should we sort the leaves?
+			// detach_hashes.sort_by(|a, b| Keccak256::hash(a).cmp(&Keccak256::hash(b)));
```
sinkingsugar: can you check actual literature? it might be good?

pallets/detach/src/lib.rs
```diff
@@ -610,7 +679,8 @@ pub mod pallet {
 					}.to_big_endian(&mut chain_id_be);
 
 					let payload = [
-						&request.hash.get_signable_hash()[..],
+						&request.collection.get_type()[..],
```
sinkingsugar: use an enum

sinkingsugar (changes requested):

pallets/detach/src/lib.rs
aang114: It makes sense since then the **user won't have to care about the order of the leaves** (although btw in the emitted event `CollectionDetached` we show the order). 

So it's not actually necessary or better, but can be more convenient for the user. **I guess I'll just implement it then...**

pallets/detach/src/lib.rs
aang114: done

pallets/detach/src/lib.rs
aang114: done

aang114:
Hi @sinkingsugar, please approve this PR (if okay) so that it can be merged. Thanks

pallets/detach/src/lib.rs
```diff
@@ -633,9 +644,9 @@ pub mod pallet {
 		}
 
 		fn get_merkle_root(detach_collection: &DetachCollection) -> Hash256 {
-			let detach_hashes = detach_collection.get_abi_encoded_hashes();
+			let mut detach_hashes = detach_collection.get_abi_encoded_hashes();
 			// TODO Review - Should we sort the leaves?
```
sinkingsugar: remove?

pallets/detach/src/lib.rs
```diff
@@ -679,7 +690,10 @@ pub mod pallet {
 					}.to_big_endian(&mut chain_id_be);
 
 					let payload = [
-						&request.collection.get_type()[..],
+						// TODO Review - In `CollectionFactory.sol` - should we have converted the enum `CollectionType` to `uint8` or `uint64`/`uint256`?
+						// In Fragnova's Smart Contract `CollectionFactory.sol`, we convert the enum `CollectionType` to `uint8` when verifying the signature.
+						// Therefore, when constructing the signature - we are also converting it to `u8`.
+						&(request.collection.get_type() as u8).to_be_bytes()[..],
```
sinkingsugar, line 694: uint8 is fine

sinkingsugar (changes requested):

sinkingsugar (approved):

PR #246, opened 2023-01-04
Pallet aliases

alebaffa:


pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+mod dummy_data;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+
+/// Enum that indicates the types of target assets linked to an alias
+#[derive(Encode, Decode, Clone, scale_info::TypeInfo, Debug, PartialEq)]
+pub enum LinkTarget<TAccountId> {
+	Proto(Hash256),
+	Fragment { definition_hash: Vec<u8>, edition_id: InstanceUnit, copy_id: InstanceUnit }, // struct variant allows a nicer UX on polkadotJS
```
sinkingsugar: Why `Vec<u8>` here?

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+		+ pallet_clusters::Config
+	{
+		/// Because this pallet emits events, it depends on the runtime's definition of an event.
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// The price (in native NOVA) for a Namespace
+		#[pallet::constant]
+		type NamespacePrice: Get<u128>;
+
+		/// The max size of namespace string
```
sinkingsugar: This should be namespace and name , both elements need a limit

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+	/// **StorageDoubleMap** that maps a (**Namespace**, **Alias** name index) with a **LinkTarget**.
+	#[pallet::storage]
+	pub type Aliases<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		Vec<u8>, // namespace
+		Twox64Concat,
+		Compact<u64>, // alias name index
+		LinkTarget<T::AccountId>, // target asset
+	>;
```
sinkingsugar: I think this looks good 👍

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+
+			let namespace = namespace.into_inner();
+
+			// Check that the namespace does not exist already
+			ensure!(!<Namespaces<T>>::contains_key(&namespace), Error::<T>::NamespaceAlreadyExists);
+
+			// burn NOVA from account's balance. The amount is set in Config.
+			let amount: <T as pallet_balances::Config>::Balance =
+				<T as Config>::NamespacePrice::get().saturated_into();
+			<pallet_balances::Pallet<T> as fungible::Mutate<T::AccountId>>::burn_from(
```
sinkingsugar: I'm not sure if burn is correct, I think it's `withdraw`, this should be just like a transaction basically, feel free to dig about it.

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+			Ok(())
+		}
+
+		/// Create an **Alias** that links to a target asset.
+		///
+		/// Only the owner of the asset can create its alias.
+		///
+		/// - `namespace`: namespace to create
+		#[pallet::weight(10_000 + T::DbWeight::get().writes(1))]
+		pub fn create_alias(
```
sinkingsugar, line 260: add `create_root_alias`? or we scrap that feature?

pallets/aliases/src/lib.rs
alebaffa: Because it is how it is set in `GetInstanceOwnerParams` struct in `pallet-fragments`, which I am reusing here to check the ownership of a fragment instance (by calling the related function in pallet-fragments).

pallets/aliases/src/lib.rs
alebaffa: Updated the comment as this is used for both.

pallets/aliases/src/lib.rs
alebaffa: I made a `update_alias_target`, callable only from root, which is what you called `update_root_alias` in the issue #226 . This function allows the root user to modify the underline `LinkTarget`.

You mean to implement a `create_root_alias` function that would create an alias without checking for the ownership of the related `LinkTarget` asset (since root has not an `AccountId` that we can retrieve, so if we want to check ownership of the asset I would need to pass a `T::AccountId` as a parameter)?

pallets/aliases/src/lib.rs
alebaffa: Right, it's withdraw to reduce account's balance. Burn would reduce the issuance. Thanks!

pallets/aliases/src/lib.rs
sinkingsugar: I just want the ability for root to create **No Namespace** Aliases.

pallets/aliases/src/lib.rs
sinkingsugar: `GetInstanceOwnerParams` is used in RPCs, should be unrelated with storage? You can translate types when going into RPCs world.

sinkingsugar (changes requested):
Nearly there, just 2 comments to sort out.

pallets/aliases/src/lib.rs
alebaffa: It is unrelated to storage, I do not store this struct in `pallet-aliases`. Since here I need to check ownership of the Fragment instance when creating the alias, reusing this struct available already in `pallet-fragment` made it easy to reuse also the same function from `pallet-fragment` that does exactly what I need here. Indeed, in this PR I also quickly refactored the function `get_instance_owner` called by the RPC in `pallet-fragment` , where I extracted the part that makes the ownership check in another function called `get_instance_owner_account_id` that returns an `AccountId`. The RPC function `get_instance_owner` calls that function and it does what it was doing before, i.e. encoding it to return a `Vec<u8>`. This way, we also have a reusable internal function in `pallet-fragment` that can be reused by other pallets if we need to check ownership of Fragment instance.

So, `GetInstanceOwnerParams` struct here is needed to just call `get_instance_owner_account_id` from `pallet-fragment`.
Can this work?

pallets/aliases/src/lib.rs
alebaffa: Creating an alias without related namespace would mean creating an entry in the `Aliases StorageDoubleMap` without key1 (key1 = namespace, key2 = alias). 
So, the options here are to either insert root aliases in that StorageDoubleMap with a "fake" namespace, which would indicate that they are "root" aliases, or create another storage dedicated to root aliases.
What do you prefer?

Also, two considerations/questions:
1. this would also mean that root aliases without namespace will be unique
2. what would be the scope of this root aliases? Can they be assigned to a namespace in the future?

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+
+	/// **StorageDoubleMap** that maps a (**Namespace**, **Alias** name index) with a **LinkTarget**.
+	#[pallet::storage]
+	pub type Aliases<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		Vec<u8>, // namespace
+		Twox64Concat,
+		Compact<u64>,             // alias name index
+		LinkTarget<T::AccountId>, // target asset
```
sinkingsugar, line 121: @alebaffa you are using the enum with the wrong `Vec<u8>` here.... how can you say its not related to storage? I'm confused.

pallets/aliases/src/lib.rs
sinkingsugar: Really confused by the answer. I added a new comment on storage part.
A fragment is (hash128, unit, unit). That's what it should be.

pallets/fragments/src/lib.rs
```diff
@@ -1922,6 +1922,36 @@ pub mod pallet {
 			Ok(())
 
 		}
+
+		/// Query the owner of a Fragment Instance. The return type is AccountId
+		pub fn get_instance_owner_account_id(
+			params: GetInstanceOwnerParams<Vec<u8>>,
+		) -> Result<T::AccountId, Vec<u8>> {
+			let definition_hash: Hash128 = hex::decode(params.definition_hash)
```
sinkingsugar: see here also you do a conversion...

sinkingsugar: from hex???

alebaffa: I did not change this. This is how it was implemented before.

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+		///
+		/// - `who`: the AccountId
+		/// - `target`: the LinkTarget containing the asset to check ownership
+		pub fn is_target_owner(
+			who: T::AccountId,
+			target: LinkTarget<T::AccountId>,
+		) -> DispatchResult {
+			match target {
+				LinkTarget::Fragment { definition_hash, edition_id, copy_id } => {
+					let owner = pallet_fragments::Pallet::<T>::get_instance_owner_account_id(
```
sinkingsugar: I would not mix rpc calls with transactions.. can't you just read what you need from storage?

pallets/aliases/src/lib.rs
sinkingsugar: Yeah I think we can establish a proprietary namespace. Say `frag`. And add to to config. That should fix all issues. 👍

pallets/fragments/src/lib.rs
sinkingsugar: Yes and it's correct in an RPC CONTEXT! Json, hexes etc, right?

pallets/aliases/src/lib.rs
alebaffa: Done.

pallets/aliases/src/lib.rs
alebaffa: Rolled this back. Now pallet-aliases checks Fragment ownership directly from storage.

pallets/aliases/src/lib.rs
alebaffa: Done.

pallets/aliases/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+		#[pallet::constant]
+		type NamespacePrice: Get<u128>;
+
+		/// The max size of namespace and alias strings
+		#[pallet::constant]
+		type NameLimit: Get<u32>;
+
+		/// The root namespace
+		#[pallet::constant]
+		type RootNamespace: Get<Vec<u8>>;
```
sinkingsugar, line 97: wow didn't realise this was possible!

sinkingsugar (approved):

PR #253, opened 2023-01-18
Improve CI triggers

Kryptos-FR:
- don't run full CI on draft PR*
- run it when draft is "ready for review"
- run it only on push when it is to the devel branch
- run it as usual on PR synchronize (i.e. push to an opened PR)

*: Basically the `setup` job will be ignored on draft PR and thus other jobs that depend on it won't run.


alebaffa (approved):
Thanks! 👍🏻

aang114 (approved):

PR #254, opened 2023-01-19
Remove tickets

alebaffa:
Remove from `pallet-protos`, `pallet-accounts` and `genesis`.

Notes for `pallet-accounts`: 
- `link` function now only mints 20% of NOVA and 80% are vested.
- `withdraw` function only for NOVA.


sinkingsugar (approved):
👍

PR #256, opened 2023-01-19
Add cluster into Proto struct

alebaffa:


pallets/protos/src/lib.rs
```diff
@@ -426,7 +429,7 @@ pub mod pallet {
 			tags: BoundedVec<BoundedVec<u8, <T as pallet::Config>::StringLimit>, T::MaxTags>,
 			linked_asset: Option<LinkedAsset>,
 			license: UsageLicense<T::AccountId>,
-			_cluster: Option<Cluster<T::AccountId>>,
+			cluster: Option<Cluster<T::AccountId>>,
```
sinkingsugar: Err, isn't that storing the whole big structure?

sinkingsugar (changes requested):

pallets/protos/src/lib.rs
alebaffa: Changed it to store only the cluster_id, i.e. `Hash128`.

sinkingsugar: Btw, otherwise made no sense, it would copy the data and leave it into a detached state. You didn't consider that?

sinkingsugar (approved):

sinkingsugar (approved):
Now it's how it should have been 👍
The previous version was concerning though, I still don't get what was the point on storing the full struct copy in protos 🤔

alebaffa:
It was a quick code I made locally that was not intended to be pushed, I know it made no sense.

sinkingsugar:
@alebaffa **devel** was with `_cluster: Option<Cluster<T::AccountId>>,` for a while, I didn't notice though 😄 let's pay more attention or when we do an audit it will be a tragedy. I can't be always alert with reviews.

PR #257, opened 2023-01-20
fix warnings in all codebase

alebaffa:


pallets/accounts/src/tests.rs
```diff
@@ -215,118 +214,9 @@ mod unlink_tests {
 mod sync_partner_contracts_tests {
 	use super::*;
 
-	fn hardcode_expected_request_and_response(
-		state: &mut testing::OffchainState,
-		lock: Lock,
-	) -> u64 {
-		let geth_url = Some(String::from("https://www.dummywebsite.com/"));
```
sinkingsugar: wow why? are we sure we dont use this?

sinkingsugar (changes requested):

pallets/accounts/src/tests.rs
alebaffa: I put it back. It is used in a test that needs to be fixed, so ignored for the moment (and so the call to this is temporarily removed) and that is why cargo saw it as unused. I forgot about this. I will fix the error in this PR too.

sinkingsugar:
Since you used a merge commit please squash when merging 👍

pallets/accounts/src/tests.rs
alebaffa: Fixed.

sinkingsugar (approved):

PR #258, opened 2023-01-21
Removed Unnecessary Comments

aang114:


sinkingsugar (approved):
You might wanna collect each point that is not too clear and we can have a meeting to go thru them btw 👍🙏

PR #267, opened 2023-02-01
More traits fixes

sinkingsugar:


runtime/src/lib.rs
```diff
@@ -393,6 +395,10 @@ mod validation_logic {
 						return false
 					}
 
+					if trait_struct.records.len() == 0 {
```
aang114, line 398: good catch

pallets/protos/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub mod pallet {
 
 	/// **StorageMap** that maps a **Trait ID** to the **name of the Trait**
 	#[pallet::storage]
-	pub type Traits<T: Config> = StorageMap<_, Identity, Hash64, Vec<u8>, ValueQuery>;
+	pub type Traits<T: Config> = StorageMap<_, Identity, Hash64, Vec<u8>, OptionQuery>;
```
aang114, line 173: what does `OptionQuery` do as opposed to `ValueQuery` btw?

runtime/src/lib.rs
```diff
@@ -414,45 +420,56 @@ mod validation_logic {
...
-									shards_info.implementing.contains(shards_trait),
+									shards_info.implementing.contains(trait_hash),
 								_ => false,
 							}
 						} else {
 							false
 						}
-					})
+					}) {
+						if let Some(_) = pallet_protos::Traits::<Runtime>::get(trait_hash) {
```
aang114: isn't this check redundant?

If one of the references has the `trait_hash` in its "implementing" field, obviously it exists, no?

aang114 (changes requested):
There is one redundant check

pallets/protos/src/lib.rs
sinkingsugar: `get` returns `Option<_>` instead of default

runtime/src/lib.rs
sinkingsugar: true

scripts/prometheus.yaml
```diff
@@ -0,0 +1,12 @@
+global:
```
aang114, line 1: what's prometheus.yaml for?

aang114 (approved):
LGTM

scripts/prometheus.yaml
sinkingsugar: to run prometheus

PR #268, opened 2023-02-02
Add Chain Extension for the Fragnova Blockchain

aang114:
This PR adds the chain extension methods defined in https://github.com/fragcolor-xyz/fragnova-chain-extension-methods/pull/1 to our runtime


aang114:
> This PR adds the chain extension methods defined in [fragcolor-xyz/fragnova-chain-extension-methods#1](https://github.com/fragcolor-xyz/fragnova-chain-extension-methods/pull/1) to our runtime

Please approve (and merge) https://github.com/fragcolor-xyz/fragnova-chain-extension-methods/pull/1, https://github.com/fragcolor-xyz/protos/pull/8 and https://github.com/fragcolor-xyz/substrate/pull/8 before approving this PR - since this PR uses those PR branches as dependencies. Thanks

aang114:
@sinkingsugar - the CI passes now. Please approve this PR if okay so that it can be merged. Thanks

runtime/src/lib.rs
```diff
@@ -935,7 +937,7 @@ parameter_types! {
 	/// Not strictly enforced, but used for weight estimation.
 	pub const MaxLocks: u32 = 50;
 	/// TODO: Documentation
-	pub const IsTransferable: bool = false;
+	pub const IsTransferable: bool = true; // TODO Review - Change this back to `false` once we update our substrate dependency
```
sinkingsugar, line 940: so you wanna merge this first then fix substrate?

runtime/src/chain_extension/tests.rs
```diff
@@ -0,0 +1,367 @@
...
+			assert_eq!(
+				contract_result.result.unwrap().data.0,
+				pallet_fragments::Inventory::<Test>::get(ALICE, definition_hash).unwrap().encode()
+			);
+
+		});
+	}
+
+	#[test]
+	pub fn give_instance_should_work() {
```
sinkingsugar, line 310: add a should not work if not authorized and such

sinkingsugar (changes requested):

runtime/src/lib.rs
aang114: yeah actually it makes more sense to fix substrate first, since we don't want `IsTransferable` to ever be `true`

sinkingsugar: it does not matter for `devel` as long as substrate fix comes next though.

runtime/src/chain_extension/tests.rs
aang114: > add a should not work if not authorized and such

added

runtime/src/lib.rs
aang114: > it does not matter for `devel` as long as substrate fix comes next though.

okay let's do that then. otherwise, if I merge the substrate PR first - there might be merge conflicts

sinkingsugar (approved):

PR #269, opened 2023-02-03
Fix cluster role settings

alebaffa:
Allow the user to provide a cluster role setting name as a normal string.

### DIFF
--- a/pallets/clusters/src/lib.rs
+++ b/pallets/clusters/src/lib.rs
@@ -11,7 +11,7 @@
 extern crate core;
 
 use codec::{Compact, Decode, Encode};
-use frame_support::{BoundedVec, transactional};
+use frame_support::{transactional, BoundedVec};
 pub use pallet::*;
 use sp_fragnova::Hash128;
 use sp_std::{vec, vec::Vec};
@@ -252,7 +252,8 @@ pub mod pallet {
 			ensure!(!<Clusters<T>>::contains_key(&cluster_id), Error::<T>::ClusterExists);
 
 			let cluster_name = name.into_inner();
-			let name_index = Self::take_name_index(&cluster_name);
+			let name_index =
+				Self::take_name_index(&cluster_name, true).expect("adding if none is true, qed");
 
 			// At creation there are no roles and no members assigned to the cluster
 			let cluster =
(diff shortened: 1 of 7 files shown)


pallets/clusters/src/lib.rs
```diff
@@ -31,9 +31,18 @@ mod benchmarking;
...
+	/// Index of the name of the role setting to be stored
 	pub name: Compact<u64>,
-	/// The data associated with the Role
+	/// The data associated with the Role to be stored
+	pub data: Vec<u8>,
+}
+
+/// The **settings** of a **Role**
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, scale_info::TypeInfo)]
+pub struct Setting {
```
sinkingsugar: Can we call this `Setting` and the current `RoleSetting` `CompactSetting` or something to avoid confusion?
Right know it's very confusing.

pallets/clusters/src/lib.rs
```diff
@@ -441,10 +458,32 @@ pub mod pallet {
...
+
+			// ensure that the Role does not contain already any of the setting names provided
+			for setting in &role_settings {
+				ensure!(
+					!<Roles<T>>::get(&cluster_id, &name_index)
+						.ok_or(Error::<T>::SystematicFailure)?
+						.settings
+						.into_iter()
+						.map(|setting| setting.name)
+						.collect::<Vec<Compact<u64>>>()
```
sinkingsugar: I'm sure you don't need to collect... just filter directly using `any` or `all`

sinkingsugar (changes requested):

pallets/clusters/src/lib.rs
sinkingsugar: without considering that this is called get on same role repeatedly in the loop 🤦‍♂️

sinkingsugar: also it seems there is no way to just change a setting currently. need to remove first.

sinkingsugar:
`T::DataLimit` still not used... not complete yet.

sinkingsugar:
@alebaffa TODO add and fix tests, I had enough of this for today 😱

PR #279, opened 2023-03-06
When uploading a shards script, the Implemented Traits must be referenced as protos 

aang114:


runtime/src/lib.rs
```diff
@@ -504,11 +505,9 @@ mod validation_logic {
...
 				let implementing = &shards_script_info_struct.implementing;
 				let implement_check = implementing.iter().all(|trait_hash| {
-					if let Some(_) = pallet_protos::Traits::<Runtime>::get(trait_hash) {
-						true
-					} else {
-						false
-					}
+					pallet_protos::Traits::<Runtime>::contains_key(trait_hash)
+						&&
+						proto_references.iter().any(|proto_hash| twox_64(proto_hash) == *trait_hash)
```
aang114, line 510: @sinkingsugar  - btw  I noticed that we check if the references actually exist in `protos.upload()` itself, **not in the validation phase**. 

**Should we check if they exist in the validation phase itself?**

sinkingsugar (approved):

PR #282, opened 2023-03-10
Ensure a Shard Script's Traits are valid in `protos.upload()` itself and not prior in the validation phase

aang114:


runtime/src/lib.rs
```diff
@@ -485,44 +476,7 @@ mod validation_logic {
...
-						return false;
-					};
-					proto_references.iter().any(|proto_hash| {
-						trait_protos.contains(proto_hash)
-					})
-				});
-
-				require_check && implement_check
-			},
+			Categories::Shards(shards_script_info_struct) => shards_script_info_struct.format == ShardsFormat::Edn,
```
aang114: @sinkingsugar  - should we validate that magic number of an edn file 😅 ?

sinkingsugar: there is no such thing, but should just check it's valid utf8

pallets/protos/src/tests.rs
```diff
@@ -643,215 +643,215 @@ mod get_protos_tests {
 		});
 	}
 
-	#[test]
```
sinkingsugar, line 646: whats going on here? why remove tests?
Should in fact add tests for the above change.

sinkingsugar (changes requested):

sinkingsugar (approved):

